Issue #182, opened 2013-03-12
Future calls to dead actors can raise Celluloid::Mailbox::DeadRecipient


michaelgpearce:
We are occasionally getting an error "dead recipient" from /celluloid-0.11.1/lib/celluloid/mailbox.rb:26:in `<<' inside of a Rails app.  This is occurring in the middle of processing a request.  I'm unsure the exact cause since we do nothing to explicitly shut down celluloid.  My guess is that this is occurring when we restart a web server in our server pool and there is a race condition between shutting down the server instance and celluloid's Kernel.at_exit hook.  Any other thoughts or ideas?

Partial strack trace:

```
/home/bkr/.bundle/4218576002/614614351/ruby/1.9.1/gems/celluloid-0.11.1/lib/celluloid/mailbox.rb:26:in `<<'
/home/bkr/.bundle/4218576002/614614351/ruby/1.9.1/gems/celluloid-0.11.1/lib/celluloid/future.rb:35:in `execute'
/home/bkr/.bundle/4218576002/614614351/ruby/1.9.1/gems/celluloid-0.11.1/lib/celluloid/actor.rb:91:in `future'
/home/bkr/.bundle/4218576002/614614351/ruby/1.9.1/gems/celluloid-0.11.1/lib/celluloid/actor_proxy.rb:62:in `future'
/home/bkr/.bundle/4218576002/614614351/ruby/1.9.1/gems/parallizer-0.4.1/lib/parallizer.rb:57:in `add_call'
/home/bkr/.bundle/4218576002/614614351/ruby/1.9.1/gems/parallizer-0.4.1/lib/parallizer.rb:42:in `block in add'
/home/bkr/.bundle/4218576002/614614351/ruby/1.9.1/gems/parallizer-0.4.1/lib/parallizer/method_call_notifier.rb:11:in `call'
/home/bkr/.bundle/4218576002/614614351/ruby/1.9.1/gems/parallizer-0.4.1/lib/parallizer/method_call_notifier.rb:11:in `method_missing'
/var/www/cart/releases/20130309032059/app/models/namespaced/price_management_system/pricing_web_service/adapter.rb:93:in `block in get_jsons_hash_from_pws'
/var/www/cart/releases/20130309032059/app/models/namespaced/price_management_system/pricing_web_service/adapter.rb:91:in `each_slice'
/var/www/cart/releases/20130309032059/app/models/namespaced/price_management_system/pricing_web_service/adapter.rb:91:in `with_index'
[...]

PR #245, opened 2013-05-10
Invalidate the block on call completion

halorgium:
When calling a method on an `Actor` from a normal `Thread`, if you invoke a block after the `Call` has completed, you will encounter a deadlock. 

This is due to the fact the sending `Thread` is no longer waiting for the block invocations. 
[gist](https://gist.github.com/halorgium/5554085/ffd1a4b4f57394d0514f486517e2a4646e66f273)

If you are calling from an Actor, this behaviour works. 
[gist](https://gist.github.com/halorgium/5554085/db96cf50b7f120b22c03437d53e6d7434d1824f0)

This change invalidates the block for both scenarios. 
We can allow block invocations in the actor case by only invalidating if the sender is a normal `Thread`. 
- [ ] Specs
- [ ] Remove extra FIXMEs
- [ ] Consider async/future + blocks with respect to invalidation

### DIFF
--- a/lib/celluloid/calls.rb
+++ b/lib/celluloid/calls.rb
@@ -22,7 +22,10 @@ module Celluloid
 
     def dispatch(obj)
       _block = @block && @block.to_proc
-      obj.public_send(@method, *@arguments, &_block)
+      result = obj.public_send(@method, *@arguments, &_block)
+      # FIXME: we should only invalidate if the sender is not an actor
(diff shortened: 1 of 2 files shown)


digitalextremist:
@niamster do you feel like this could be part of the same thinking behind resolving #182?

tarcieri:

PR #282, opened 2013-06-18
Nicer logging for debugging

halorgium:
It is almost impossible to determine where logs after coming from unless you add metadata yourself. 
We have this information inside the system, so this exposes that outward. 

Quite verbose, but useful.

### DIFF
--- a/lib/celluloid.rb
+++ b/lib/celluloid.rb
@@ -14,7 +14,7 @@ module Celluloid
 
   class << self
     attr_writer   :actor_system     # Default Actor System
-    attr_accessor :logger           # Thread-safe logger class
+    attr_reader   :logger           # Thread-safe logger class
     attr_accessor :task_class       # Default task type to use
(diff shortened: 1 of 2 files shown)


digitalextremist:
@e2 is this mergeable now you think? Since 2013 it seems there has been BSD progress.

@halorgium would you mind looking at this and your stale branches for `0.18.0`?

e2:
I saw this PR years ago.

Celluloid needs a better debugging framework, but I'm sure that would hurt performance too much - because the areas you'd want the information are pretty critical, and the log files aren't really useful any more (too much stuff going on in there).

It's partially why I suggested: https://github.com/celluloid/celluloid/issues/548.

The best solution is to probably set Celluloid up so it can traces externally (without a performance hit). Kind of like a strace().

This probably needs a separate issue though.

tarcieri:

PR #121, opened 2013-10-06
Problems trapping signals: deadlock on 1.9, impossible on 2.0

halorgium:
I'll try take a look at this tomorrow if I get time.

`ruby 1.9.3p286 (2012-10-12 revision 37165) [x86_64-darwin12.2.1]`
`Celluloid 0.12.3`

``` ruby
class Foo
  include Celluloid

  def bar
    sleep 0.1
  end
end

pool = Foo.pool(size: 2)
Signal.trap :HUP, Proc.new { puts pool.size }
puts Process.pid
200.times { pool.async.bar }
```

In another terminal, spam the process with a bunch of HUPs. After ~10 seconds the puts will print, followed by the process exiting (as if I wasn't trapping the signal ???) followed by a deadlock trace.

The deadlock itself is perhaps not so weird, what is weird is that the HUPs appear to cause the process to exit. If you change the HUP block to do something non-celluloid related, then it behaves as expected.

Output: https://gist.github.com/4102620

### DIFF
--- a/lib/celluloid.rb
+++ b/lib/celluloid.rb
@@ -131,6 +131,20 @@ module Celluloid
       actor_system && actor_system.running?
     end
 
+    def trap(signal)
+      r, w = IO.pipe
+
+      Thread.new {
+        while r.read(1)
+          yield
+        end
+      }
+       
+      Signal.trap(signal) {
+        w.write(".")
+      }
+    end
+
     def register_shutdown
       return if @shutdown_registered
       # Terminate all actors at exit
--- a/spec/celluloid/trap_spec.rb
+++ b/spec/celluloid/trap_spec.rb
@@ -10,16 +10,33 @@ class Bikeshed
   end
 end
 
-describe "Celluloid.trap" do
-  it "handles INFO" do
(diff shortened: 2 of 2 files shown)


marshall-lee:
Hello!

I tried to extract this approach seen many times before into a gem.
What I have got: https://github.com/marshall-lee/tricky_signals

ioquatix:
I made a very simple notification pipe for cross-fork notifications: 

https://github.com/ioquatix/process-daemon/blob/master/lib/process/daemon/notification.rb

I'm not sure but this could easily form the basis for a celluloid implementation of the same technique. It works well.

ioquatix:
@marshall-lee I like your gem, it's interesting.

tarcieri:

thesmart:
@tarcieri I'm still running into this issue: https://github.com/slack-ruby/slack-ruby-bot/issues/116

Was a signal handling API ever merged?  It seems that signal handling breaks in Celluloid when running ruby 2.3.

tarcieri:
There's a solution in previous versions of Sidekiq that still utilized Celluloid you can look at. There is not presently a standard API for this.

PR #350, opened 2013-10-12
Add shortcut to check if actor is dead

cpuguy83:
Keep finding I want to check if something is dead rather than alive.
Just a little shortcut for that.


digitalextremist:
This seems to be outdated. I will break it out into another branch.

digitalextremist:
Yeah, tests are failing. This is moved over into a new PR #522 now, ready to be updated over there.

digitalextremist:
Fixed `rspec` syntax and it's all gone now.

PR #371, opened 2014-01-27
Pooled fibers

cheald:
(Recreated based off of master now)

This adds support for a variant on TaskFiber which pools fibers for a given thread. This is primarily a triage patch for JRuby 1.7.5+ which isn't currently pooling the threads that back its fibers, resulting in a ridonkulous amount of CPU burn as it spins up a new thread per Fiber. See jruby/jruby#1443 for full details there.

This is the same fiber pattern used by rack-fiber_pool, among others. In my tests, it's completely eliminated the JRuby CPU thrashing, and actually improved performance of Celluloid under MRI, too.

Using [this very unrigorous benchmark](https://gist.github.com/cheald/94097ddb2a855cdb2583), I saw the following results:

|              | TaskFiber           | TaskPooledFiber |
|--------------|---------------------|-----------------|
| 2.0.0-p247   | 4.08 sec            | 2.79 sec        |
| jruby-1.7.10 | 29.5 sec (not a typo) | 5.04 sec        |

### DIFF
--- a/lib/celluloid/tasks.rb
+++ b/lib/celluloid/tasks.rb
@@ -152,3 +152,4 @@ end
 
 require 'celluloid/tasks/task_fiber'
 require 'celluloid/tasks/task_thread'
+require 'celluloid/tasks/task_pooled_fiber'
--- a/lib/celluloid/tasks/task_pooled_fiber.rb
+++ b/lib/celluloid/tasks/task_pooled_fiber.rb
@@ -1,0 +1,56 @@
+module Celluloid
+
+  class FiberPool
+    def acquire(&block)
+      # If the fiber pool has grown too much, shift off and discard
+      # extra fibers so they may be GC'd. This isn't ideal, but it
+      # should guard against runaway resource consumption.
+      fiber_pool.shift while fiber_pool.length > 10
+      fiber = fiber_pool.shift || create_fiber
+      fiber.resume(block)
+      fiber
+    end
+
+    def create_fiber
+      fiber_pool = self.fiber_pool
+      Fiber.new do |blk|
+        loop do
+          # At this point, we have a pooled fiber ready for Celluloid to call #resume on.
+          Fiber.yield
+
+          # Once Celluloid resumes the fiber, we need to execute the block the task was
+          # created with. This may call suspend/resume; that's fine.
+          blk.call
+
+          # Once Celluloid completes the task, we release this Fiber back to the pool
+          fiber_pool << Fiber.current
+
+          # ...and yield for the next #acquire call.
+          blk = Fiber.yield
+        end
+      end
+    end
+
+    # Fiber pool for this thread. Fibers can't cross threads, so we have to maintain a
+    # pool per thread.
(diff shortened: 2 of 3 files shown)


digitalextremist:
@cheald not sure what happened here... why this collected dust. Are you still using this code yourself? I will factor this into #512

cheald:
We are still using it with great success, yes.

digitalextremist:
I am very excited to test this myself @cheald and bring it in.

digitalextremist:
So far, this is amazing. I need to test it further but in my extremely complicated implementation it seems like it's removing issues I've been stuck with congestion-wise for *years*.

digitalextremist:
@tarcieri I think this is a separate gem -- `celluloid-taskpooledfiber`

tarcieri:
I'd like to get pools out of core entirely FWIW

digitalextremist:
Agreed, #512 -- I'm saying this is *another* pool oriented gem.

tarcieri:
I'm all for multiple competing pool gems :wink:

digitalextremist:
Right on. I have it coexisting well with the existing pool implementation, but it's different enough to where I wouldn't want to lump it into `celluloid-pool`

digitalextremist:
That looks good also ( `celluloid-coordinator` ) -- I'll bring that over also. And I think I may have another myself that I call `spans` but that can hold a while probably.

digitalextremist:
Yeah, #478 is looking really good right about now.

cheald:
Sorry, I've been out all evening. Yes, this is simply an alternate Task type. If it's desired, I can publish it as a gem; shouldn't be too hard.

digitalextremist:
Dinged you on LinkedIn in on the topic, I was going to ask if you'd mind being attributed and I'd cut the gem for you either stand-alone or in `celluloid-extras` so it can be actively maintained and kept current with changes as they happen in the API ( unless you are prepared to do that ). There are several pieces of Celluloid being decoupled so it'd be in good company.

cheald:
Oh, that's fine. I just pushed https://github.com/cheald/celluloid-task-pooled-fiber but I'm happy to nuke it if it's already in motion to be made available somewhere.

cheald:
BTW, if you want, I'm happy to just give you the commit bit on that repo, too. Whatever's easiest - we're definitely going to be using TaskPooledFiber for a long while, so I will have a vested interest in keeping it maintained. :)

cheald:
I went ahead and pushed https://rubygems.org/gems/celluloid-task-pooled-fiber - if this ends up getting rolled into another gem I'll just pull it.

digitalextremist:
The way you've got it now will work out totally fine. I look forward to maintaining this together. I'm very impressed with its performance in my use so far. My own platform is like new, snappier than ever before. Glad this code came back to the foreground. Excited to see it widely used.

digitalextremist:
This will be well taken care of, with your cooperation or approval of any decisions, in #549 once the new API changes land.

PR #382, opened 2014-02-04
added spec which specifies the issue of reactor crashing when Kernel.tim...

TiagoCardoso1983:
...eout blows (this spec fails)

This merge request is related to the issue from: https://github.com/celluloid/celluloid-io/pull/98#issuecomment-34047478


digitalextremist:
Fixed in 9726a4c

PR #384, opened 2014-02-09
Add bundler as development_dependency

vjoel:
For those of us who don't always have bundler installed...


digitalextremist:
`bundler` got added already, in the year+ that this has been sitting. So I added `coveralls` in a3b4d6b

PR #385, opened 2014-02-10
Add fluent API for #async

vjoel:
Does this make sense? The first thing I tried while learning #async was to chain calls, but it didn't work because of the nil return value. See [examples/stack.rb](examples/stack.rb). (Are my comments there correct?)


digitalextremist:
Merging and patching example with slight/safe API behavioral tweak.

PR #387, opened 2014-02-13
Remove actor proxy

jnicklas:
ActorProxy isn't really used as a proxy anymore, and there doesn't seem to be any reason to wrap the actor in an actor proxy. Instead we can just send through the whole actor, which is simpler.


ioquatix:
Bump :)

tarcieri:
Original point still stands: the intention of `ActorProxy` was to enable non-Cell `Actors` even if it isn't otherwise "used".

There still may be work involved in making that practical/usable, but I'd rather not backpedal by hastily removing it because it isn't fully baked.

tarcieri:
Closing as stale

PR #396, opened 2014-03-07
Using timers instead of Time.now in Mailbox and EventedMailbox

jstoja:
Reading the code and found an easy todo. I hope it's well implemented, if not, I'll correct.


digitalextremist:
This is already implemented, and with the new `Timers` api.

PR #399, opened 2014-03-11
An actor should know who its supervisor is.

Senjai:
This is my attempt, feedback welcome. Not too cheery about the test, but
there was no better place to put it.


digitalextremist:
This seems relevant to #506 and #509. I'd like to include some/all of it in `refactor-supervision`

digitalextremist:
I will address this in #511

PR #478, opened 2014-11-29
Omnirepo

tarcieri:
As originally discussed here:

https://groups.google.com/d/msg/celluloid-ruby/SsGjMAtYacU/09tY4ghCd6gJ

This merges the following projects into an omnibus repository:

* Celluloid
* Celluloid::IO
* Celluloid::ZMQ

All commit history is preserved.

Still to consider:

* DCell
* Timers


digitalextremist:
I remember a lot of talk about this. @tarcieri, is this abandoned? Out of date? Still going?

tarcieri:
I'm kind of torn on if this is a good idea or not. If you'd like to pick it up, I'd suggest basing it off the latest versions of all of the libraries.

It probably deserves more discussion.

digitalextremist:
Roger. I will bring this up during the discussion leading to `0.17.0`

digitalextremist:
Maybe we can pause on this altogether until we're ready to make the jump to `1.0` actually, to focus completely on that, then make repository changes?

ioquatix:
The reason to keep things separate is because it allows you to update them independently. If timers was part of celluloid, you couldn't necessarily update one without updating the other.

tarcieri:
@ioquatix this isn't about combining the gems. It's about having a single git repository, so when there are cross-cutting concerns across libraries, you can update them all in one shot.

digitalextremist:
I'm going to redo this today @tarcieri, then we can make the call on whether to commit it.

digitalextremist:
Abandoning this version, but using it for reference.

PR #485, opened 2015-01-03
Notification fanout support for symbols

### LINKED ISSUE #482: Notifications don't work with a symbol topic
Notifications are never received if the topic is a symbol. Since apparently a String is expected, should we be coercing the topic to a string? Or should it work for symbols and other objects?

For now, I changed the [Notifications wiki](https://github.com/celluloid/celluloid/wiki/Notifications) to reference a "topic string" instead.

### Works

```ruby
class Subscriber
  include Celluloid
  include Celluloid::Notifications

  def initialize
    subscribe('topic', :handler)
  end

  def handler(topic, *args)
    puts "notified of #{topic}"
  end
end

class Publisher
  include Celluloid
  include Celluloid::Notifications

  def notify
    publish('topic')
  end
end

Subscriber.new
Publisher.new.notify
```
### Fails

The handler is never called.

```ruby
class Subscriber
  include Celluloid
  include Celluloid::Notifications

  def initialize
    subscribe(:topic, :handler)
  end

  def handler(topic, *args)
    puts "notified of #{topic}"
  end
end

class Publisher
  include Celluloid
  include Celluloid::Notifications

  def notify
    publish(:topic)
  end
end

Subscriber.new
Publisher.new.notify
```

sb8244:
Because of the .to_s comparison, symbols were not able to be compared

Fixes #482

### DIFF
--- a/lib/celluloid/notifications.rb
+++ b/lib/celluloid/notifications.rb
@@ -71,7 +71,7 @@ module Celluloid
       end
 
       def subscribed_to?(pattern)
-        !pattern || @pattern === pattern.to_s
+        !pattern || @pattern === pattern.to_s || @pattern === pattern
       end
 
       def matches?(subscriber_or_pattern)
--- a/spec/celluloid/notifications_spec.rb
+++ b/spec/celluloid/notifications_spec.rb
@@ -19,8 +19,8 @@ describe Celluloid::Notifications, actor_system: :global do
     include Celluloid
     include Celluloid::Notifications
 
-    def die
-      publish("death", "Mr. President")
+    def die(topic = "death")
+      publish(topic, "Mr. President")
     end
   end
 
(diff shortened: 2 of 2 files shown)


tarcieri:
Can you confirm this behaves the same way as `ActiveSupport::Notifications`?

sb8244:
It looks like `AS::Notifications` does it even simpler [fanout.rb#L106](https://github.com/rails/rails/blob/a8439bc470e90cce44fef6c182d94cf01a944f1e/activesupport/lib/active_support/notifications/fanout.rb#L106):

```ruby
def subscribed_to?(name)
  @pattern === name
 end
```

The simplicity of this is that regexp will match strings / symbols. However, a subscription to a string with a symbol emitted will not match. That is why I kept the check that uses `.to_s`. I personally don't think that the symbol/string matching is very important, but others might expect it to be handled transparently behind the scenes.

tarcieri:
I'm kind of confused by why it supports `!pattern`... but I guess this is ok for now.

PR #486, opened 2015-01-03
Make InternalPool handle Exception

### LINKED ISSUE #484: Raising Exception in pool causes mem leak
Dug into https://github.com/mperham/sidekiq/issues/2107 and found it was a Celluloid issue. The issue arises from raising an Exception in a pool'd worker. The thread cleanup never gets run here: https://github.com/celluloid/celluloid/blob/a6584e787926de193b3aa64c70f03f125e3dc8cc/lib/celluloid/internal_pool.rb#L145

Here is a gist that recreates the issue with only Celluloid https://gist.github.com/sb8244/0a7e505dd1dafe432b31

So, is this a bug that should be fixed? I would love to contribute to Celluloid and can take a shot at this!

sb8244:
Fixes #484

### DIFF
--- a/spec/celluloid/internal_pool_spec.rb
+++ b/spec/celluloid/internal_pool_spec.rb
@@ -43,6 +43,7 @@ describe Celluloid::InternalPool do
       end
     end
   end
+
   it "cleans thread locals from old threads" do
     thread = subject.get { Thread.current[:foo] = :bar }
 
--- a/spec/celluloid/internal_pool_spec.rb
+++ b/spec/celluloid/internal_pool_spec.rb
@@ -22,6 +22,27 @@ describe Celluloid::InternalPool do
     subject.busy_size.should eq 0
   end
 
+  context "with errors in the threads" do
+    [StandardError, Exception].each do |exception_class|
+      it "puts error'd threads back into the pool" do
+        subject.idle_size.should be_zero
+        subject.busy_size.should be_zero
+
+        queue = Queue.new
+        Celluloid::Logger.should_receive(:crash).once
+        subject.get { raise exception_class.new("Error") }
+
+        subject.idle_size.should be_zero
+        subject.busy_size.should eq 1
+
+        queue << nil
+        sleep 0.01 # hax
+
+        subject.idle_size.should eq 1
+        subject.busy_size.should eq 0
+      end
+    end
+  end
   it "cleans thread locals from old threads" do
     thread = subject.get { Thread.current[:foo] = :bar }
 
--- a/spec/celluloid/internal_pool_spec.rb
+++ b/spec/celluloid/internal_pool_spec.rb
@@ -29,7 +29,7 @@ describe Celluloid::InternalPool do
         subject.busy_size.should be_zero
 
         queue = Queue.new
-        Celluloid::Logger.should_receive(:crash).once
+
         subject.get { raise exception_class.new("Error") }
 
         subject.idle_size.should be_zero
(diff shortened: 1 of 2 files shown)


tarcieri:
Might be good to check the history of `InternalPool` to make sure we're not retracing our steps in the sand...

tarcieri:
At the very least, it should rethrow the `Exception` when it's done. I'm worried about the intervening code having issues, though...

sb8244:
If it rethrow's an exception when it's done, the `put thread` line will not run.

Here is the commit that appears to have changed this from `rescue Exception` to just `rescue` https://github.com/sb8244/celluloid/commit/6c3798533fe9384340eb848edff91ff5cd90ff3b. However, I also see that the `put` used to occur in an ensure block. One solution that passes my local tests is to keep this old ensure behavior:

I'm not sure if re-raising the exception is the best course of action or not, as I worry that raising it could cause unexpected errors else-where. Although, none of the specs fail when raising it.

```ruby
      thread = Thread.new do
        while proc = queue.pop
          begin
            proc.call
          rescue Exception => ex
            Logger.crash("thread crashed", ex)
            raise
          ensure
            put thread
          end
        end
      end
```

tarcieri:
Okay, per 6c37985 we are definitely retracing our steps in the sand here. I can't find my specific justification for that change.

What exactly is leaking? It sounds like we need to do some sort of cleanup for `Exception`, but I don't think what you're proposing is it.

We should definitely rethrow `Exception` rather than swallowing it. It's important to many Ruby functions, including handling `Interrupt` (i.e. Ctrl-C)

sb8244:
The leak is that `put thread` is not being called and `@all_threads` keeps every old actor reference that died by raising `Exception`. I think not rescuing `Exception` but ensuring `put thread` fixes this issue gracefully.

tarcieri:
Okay, looks like `@all_threads` was added in df1a8c00 by @cheald.

The way df1a8c00 is written, I think `InternalPool` needs another method to clean up a crashing thread.

That said, I think having `@all_threads` may have deleterious performance characteristics, specifically `@all_threads.delete(thread)` will get called whenever a thread is returned to the pool and is O(1).

I think the best solution is probably to get rid of `@all_threads`.

sb8244:
`@all_threads.delete(thread)` is O(N), right? I think O(1) may be a typo.

O(N) is of course negative on performance, but `@all_threads` should never be greater than the number of running threads in the pool, which is not going to be very high. However, optimizing it away would be nice!

I don't mind giving it a shot, but I think it's a bit beyond my knowledge of Celluloid to tackle a refactor like this. I pushed up the `ensure` solution that doesn't `rescue Exception`, as I still think that is the correct behavior whether or not `@all_threads` should be there or not.

spec/celluloid/internal_pool_spec.rb
```diff
@@ -22,6 +22,27 @@
...
+        subject.busy_size.should eq 1
+
+        queue << nil
+        sleep 0.01 # hax
+
+        subject.idle_size.should eq 1
+        subject.busy_size.should eq 0
+      end
+    end
+  end
```
tarcieri: Can you add another newline below here?

tarcieri:
This seems ok for now with one caveat

spec/celluloid/internal_pool_spec.rb
sb8244: Thanks, don't know how I missed that

Issue #430, opened 2014-05-24
supervisors_and_registry.rb doesn't seem to be working correctly


ioquatix:
> ruby ./supervisors_and_registry.rb
	*** Demonstrating using the Supervisor API directly
	We should be in a clean state now: clean
	Brace yourself for a crash message...
	Uhoh, we crashed the actor...
	The supervisor should automatically restart the actor
	We should now be in a clean state again: clean
	*** Demonstrating using the actor registry
	We should be in a clean state now: clean
	Brace yourself for a crash message...
	E, [2014-05-24T17:30:08.083770 #22395] ERROR -- : Actor crashed!
	NameError: undefined local variable or method `oh_crap_im_totally_broken' for #<Celluloid::CellProxy(MyActor:0x80a2ddcc) @state=:dirty>
		./supervisors_and_registry.rb:16:in `broken_method'
		/Users/samuel/Documents/Programming/Internet/celluloid/lib/celluloid/calls.rb:26:in `public_send'
		/Users/samuel/Documents/Programming/Internet/celluloid/lib/celluloid/calls.rb:26:in `dispatch'
		/Users/samuel/Documents/Programming/Internet/celluloid/lib/celUhoh, we crashed the actor...
	The supervisor should automatically restart the actor
	luloid/calls.rb:63:in `dispatch'
		/Users/samuel/Documents/Programming/Internet/celluloid/lib/celluloid/cell.rb:60:in `block in invoke'
		/Users/samuel/Documents/Programming/Internet/celluloid/lib/celluloid/cell.rb:71:in `block in task'
		/Users/samuel/Documents/Programming/Internet/celluloid/lib/celluloid/actor.rb:362:in `block in task'
		/Users/samuel/Documents/Programming/Internet/celluloid/lib/celluloid/tasks.rb:55:in `block in initialize'
		/Users/samuel/Documents/Programming/Internet/celluloid/lib/celluloid/tasks/task_fiber.rb:15:in `block in create'
	D, [2014-05-24T17:30:08.180101 #22395] DEBUG -- : Terminating 7 actors...
	W, [2014-05-24T17:30:08.180710 #22395]  WARN -- : Terminating task: type=:exit_handler, meta={:method_name=>:restart_actor}, status=:callwait
		Celluloid::TaskFiber backtrace unavailable. Please try `Celluloid.task_class = Celluloid::TaskThread` if you need backtraces here.
[...]

PR #488, opened 2015-01-03
Fix supervisors and registry race conditions

sb8244:
Issue #430 shows that the example doesn't work out of the box. When I ran this locally, I would often get crashes on line 46 and 78 of `examples/supervisors_and_registry.rb`. The error was always a nil error when there should have been an actor. Sleeping seems to be an accepted hack when looking at the internal pool specs.

### DIFF
--- a/examples/supervisors_and_registry.rb
+++ b/examples/supervisors_and_registry.rb
@@ -17,6 +17,11 @@ class MyActor
   end
 end
 
+def ensure_actor_ready
+  # Hack to ensure crashed actor is ready before accessing the restarted actor
+  sleep 0.1
+end
+
 #
 # Using the Supervisor API directly
 #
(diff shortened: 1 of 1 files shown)


tarcieri:
This shows there's a race condition and you're working around it. I'd prefer a real solution to the race condition.

sb8244:
Could it be Ruby's GIL that is causing the race? If I crash a thread and then execute code immediately after the crash, the thread's crash handler may not get to run. The sleep allows the crash handler to run before the next instruction in the main thread.

tarcieri:
It's more like a lack of synchronization

sb8244:
I'll poke around that a bit.

sb8244:
Stumped on the synchronization here. It seems to me that it is the `SuperviserGroup#actors` call that is querying for status. `member.restart` affects the actors and appears to be where the race is occurring.  

I wrapped all blocks of code that edited `@members` in synchronization but issue is still present.

PR #496, opened 2015-02-27
Fix silently failing terminate() errors

e2:
Cell#shutdown failed while checking for a finalizer, because
PoolManager#respond_to?(meth, true) was not fully implemented.

### DIFF
--- a/lib/celluloid/pool_manager.rb
+++ b/lib/celluloid/pool_manager.rb
@@ -136,12 +136,23 @@ module Celluloid
     end
 
     def respond_to?(meth, include_private = false)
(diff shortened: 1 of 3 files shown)


e2:
@tarcieri - stubbing and checking `Celluloid::Logger.crash` seems very helpful (probably would help in other specs to detect other "silent" errors).

One build failed simply because because Travis ran out of available threads, and the PoolManager finalizer crashed. 

I'm not sure how to best fix this though.

tarcieri:
The test suite is flaky :frown:

It's passed on both Rubies at least once. Calling this good. Thanks!

PR #497, opened 2015-03-10
Update RSpec to 3.2

e2:
- convert specs to RSpec 3.2
- fix a few broken specs
- tweaks for Travis
- rspec tweaks
- rspec-retry to rerun failures on Travis

### DIFF
--- a/examples/cigarette_smokers.rb
+++ b/examples/cigarette_smokers.rb
@@ -218,7 +218,7 @@ class Waitress
 
   def note_smokers
     active_smokers = @table.smokers.select(&:smoking?)
-    puts "*** Active smokers: #{active_smokers.map(&:name).join(", ")}"
+    puts "*** Active smokers: #{active_smokers.map(&:name).join(', ')}"
     puts "!!! EVERYBODY SMOKES !!!" if active_smokers.size == @table.smokers.size
   end
 end
--- a/lib/celluloid/calls.rb
+++ b/lib/celluloid/calls.rb
@@ -132,7 +132,6 @@ module Celluloid
 
   # Asynchronous calls don't wait for a response
   class AsyncCall < Call
-
     def dispatch(obj)
       CallChain.current_id = Celluloid.uuid
       super(obj)
@@ -142,7 +141,6 @@ module Celluloid
     ensure
       CallChain.current_id = nil
     end
-
   end
 
   class BlockCall
@@ -163,5 +161,4 @@ module Celluloid
       @sender << BlockResponse.new(self, response)
     end
   end
-
 end
--- a/.rspec
+++ b/.rspec
@@ -1,5 +1,4 @@
 --color
 --format documentation
---backtrace
 --order random
 --warnings
--- a/.rspec
+++ b/.rspec
@@ -2,3 +2,4 @@
 --format documentation
 --order random
 --warnings
+--require spec_helper
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -118,11 +118,6 @@ EmptyLineBetweenDefs:
 EmptyLines:
   Enabled: false
 
-# Offense count: 8
-# Cop supports --auto-correct.
-EmptyLinesAroundBody:
-  Enabled: false
-
 Encoding:
   Enabled: false
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,4 +25,5 @@ before_install:
   # Only use 1 job until Travis fixes the rbx --jobs issue.
   - if [ "$TRAVIS_RUBY_VERSION" == "rbx-2" ] ; then export BUNDLE_JOBS=1 ; fi
 
-install: bundle install --retry=3
+sudo: false
+install: bundle install --retry=3 --without=development
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ end
 group :test do
   gem 'benchmark_suite'
   gem 'rspec', '~> 3.2'
+  gem 'rspec-retry'
 end
 
 group :gem_build_tools do
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ group :development do
   gem 'guard-rspec'
   gem 'benchmark_suite'
   gem 'rubocop'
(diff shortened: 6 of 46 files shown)


examples/cigarette_smokers.rb
```diff
@@ -218,7 +218,7 @@ def refill_table
 
   def note_smokers
     active_smokers = @table.smokers.select(&:smoking?)
-    puts "*** Active smokers: #{active_smokers.map(&:name).join(", ")}"
+    puts "*** Active smokers: #{active_smokers.map(&:name).join(', ')}"
```
digitalextremist: Curious why, here -- because we're within `"`?

digitalextremist:
@e2, it seems the bulk of this is to change the way tests are performed for the new version? Is it backward compatible with older versions?

And I notice several Celluloid internals were modified to remove space between method definitions in the class bodies. Not sure if this kind of compression is what we're looking for? What's the code standard convention here @tarcieri?

e2:
@digitalextremist - this is mostly to migrate to RSpec >= 3.0 (which was released ages ago). The files should be compatible with RSpec 2.99 (except maybe the RSpec 3 specific config), though RSpec 3.0 has better features and better checking (e.g. the mock validation could help turn many Celluloid "integration specs" into unit tests with proper mocking - and especially help with the flaky specs).

For Celluloid specifically, the difference is mostly avoiding deprecations.

Also, project with the monkey-patching disabled could not use the Celluloid RSpec helpers (because the shared examples in `lib/celluloid/rspec` depended on the monkey-patching).

> And I notice several Celluloid internals were modified to remove space between method definitions in the class bodies.

I did whatever matches RuboCop settings - if there's any specific style guide, it should be set in RuboCop's config. The compressed spaces and quotes are mostly auto-correctable by RuboCop.

(RuboCop isn't run on Travis though).

e2:
I'm not sure why the failure happened - that should've been handled by rspec-retry...

examples/cigarette_smokers.rb
e2: It's a standard RuboCop offense - mostly related to Ruby's default style guide (using single quotes instead of double quotes for constant strings).

lib/celluloid/calls.rb
```diff
@@ -132,7 +132,6 @@ def wait
 
   # Asynchronous calls don't wait for a response
   class AsyncCall < Call
-
```
digitalextremist: These are the types of code compressions I am talking about, in a few internal Celluloid classes.

tarcieri: This looks fine to me

tarcieri:
Overall this looks good, but I'm also curious why rspec-retry didn't work...

tarcieri:
Also I'm down to get the code to the state where we don't rely on monkeypatching...

e2:
I think rspec-retry ran out of retries (3 may not be enough - especially when test is rerun successively). 

I'll work something out, especially since retries are needed in development too...

tarcieri:
Feel free to mark any of the StackDumper tests that are flaking out as pending because they're flaky. They've been notoriously unreliable, and unreliable tests suck :wink:

e2:
@tarcieri - yes, I'll mark flaky tests with `flaky: true`, so only those get reruns - and so they can be easily reconfigure in the `spec_helper.rb`. (Flaky specs should be explicit, while normal tests shouldn't need reruns - it's confusing during dev).

e2:
Ok, I'm done on my side.

tarcieri:
@e2 another build failure :cry:

e2:
LOL! 5 times is not enough!

```
RSpec::Retry: 2nd try ./spec/celluloid/stack_dump_spec.rb:58
RSpec::Retry: 3rd try ./spec/celluloid/stack_dump_spec.rb:58
RSpec::Retry: 4th try ./spec/celluloid/stack_dump_spec.rb:58
RSpec::Retry: 5th try ./spec/celluloid/stack_dump_spec.rb:58
    should have the correct roles (FAILED - 1)
```

e2:
@tarcieri - any idea why it's failing? (Why the first thread is not present in the results).

tarcieri:
The entire stack dump spec is racy and not well-written. I'd say disable
all the flaky specs until someone can investigate/rewrite them.

>

-- 
Tony Arcieri

e2:
The fix looks dumb, but basically there are too many messages - and the internal_pool sometimes doesn't have a chance to end up with the 3 tested threads (2 busy + 1 idle). 

The expected idle thread (`nil`) comes from an "initialize" message - the 3rd thread for that message doesn't happen if there isn't enough time after the blocking() call (the reason for the first sleep()). 

I think the InternalPool should simply be mocked - because that's all the StackDump test is really using as a "fixture".

tarcieri:
The StackDump tests are definitely the biggest culprits in the flakiness of the test suite, and having a flaky test suite really sucks. I definitely do appreciate you doing the work to shore it up!

My vote is to disable them until they can be fixed. Flaky tests are useless and detrimental. If you have ideas to fix the great, if not mark them pending.

e2:
Overall, threads combined with mutexes and timers are hard to unit test. The only way to do that theoretically is to implement a custom "scheduler" or "virtual machine" (e.g. you can't deterministically call code after a sleep has started). 

Not worth the effort. There are some tests that can be quite elegantly turned into unit tests using queues - but that alone can be quite mind bending when combined with mocks.

This makes "integration specs" easier than trying to force Celluloid into being convenient for unit tests.

The best idea I have is to gradually make it impossible to create flaky specs in the first place - by not making the full Celluloid stack available in specs (e.g. by partially making threads, conditionals and timers unavailable during unit tests).

A first step would be to fail tests if threads are left hanging (every spec should explicitly clean up after itself). But this is probably done while fixing bugs, though.

Oh, and since RSpec 3.0 pending tests fail if they "pass" (pending tests are expected to fail, or that itself is a failure). So yeah, flaky tests are detrimental - even more so.

All MRI versions are green, btw.

tarcieri:
This looks good to me now. Thanks!

Issue #500, opened 2015-03-17
evented mailbox: the mailbox is deadlocked if reactor wakeup routine is blocked


niamster:
In celluloid/celluloid-zmq#40 it's described a situation when the reactor wakeup routine may sleep because outgoing FIFO is full.
This prevents from processing of the pending incoming messages.

niamster:
The fix was already merged.

PR #502, opened 2015-03-17
fix issue #500: do not hold mailbox lock while calling reactor wakeup routine

niamster:
A more elegant solution to avoid retaking the mutex


niamster:
@ioquatix any chance to get it merged?

ioquatix:
@tarcieri probably needs to check it.

tarcieri:
I think this should be fine...

PR #505, opened 2015-03-24
Closes #503

### LINKED ISSUE #503: Registry not working for SupervisionGroups via supervise
I have a situation where I am customizing the behavior and details of a SupervisionGroup. Using pools works just fine, but if I want to supervise only one actor, it appears as though the registry isn't working correctly. 
```
@supervisor = Celluloid::SupervisionGroup.run!

# works
@supervisor.pool(SocialData::Managers, as: :data_managers, size: 2, args: [data_name: "twitter"])
@supervisor[:data_managers].process 

# returns nil:NilClass (NoMethodError)
@supervisor.supervise(SocialData::Manager, as: :data_manager, args: [data_name: "twitter"])
@supervisor[:data_manager].process 
```
Based on everything I see, I should be able to use the above syntax to supervise one actor via a supervision group. The Celluloid::Actors approach also doesn't show the actor in the registry. Is this intentional? I am currently working with the latest from Master.

digitalextremist:
This resolves the regression in #503 with `celluloid/autostart`

@tarcieri I noticed there are Class methods declared just above the Instance method being tweaked here, and they are of the previous approach. Wondering if those should also be changed to the new method.

### DIFF
--- a/lib/celluloid/supervision_group.rb
+++ b/lib/celluloid/supervision_group.rb
@@ -33,27 +33,27 @@ module Celluloid
       end
 
       # Register an actor class or a sub-group to be launched and supervised
-      # Available options are:
-      #
-      # @option options [Symbol, Object] :as register this application in the Celluloid::Actor[] directory
-      # @option options [#call, Object] :args start the actor with the given
-      #   arguments (lazy evaluation if it responds to #call)
-      def supervise(klass, options = {})
+      def supervise(klass, *args, &block)
         blocks << lambda do |group|
-          group.add klass, options
+          group.add(klass, prepare_options(args).merge(:block => block))
+        end
+      end
+
+      def supervise_as(name, klass, *args, &block)
+        blocks << lambda do |group|
+          group.add(klass, prepare_options(args).merge(:block => block, :as => name))
         end
       end
 
       # Register a pool of actors to be launched on group startup
-      # Available options are:
-      #
-      # @option options [String, Symbol] :as register this application in the Celluloid::Actor[] directory
-      # @option options [#call, Object] :args start the actor with the given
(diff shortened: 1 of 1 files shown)


tarcieri:
I think the instance methods on SupervisionGroup shouldn't exist. They're unsafe.

digitalextremist:
@tarcieri so you would rather a potential breaking change, and to move the tweaked pieces up from the instance methods raised by #503 into the class methods, and the documentation changed to reflect that?

digitalextremist:
To be clear, we're talking about removing `pool` and `supervise` and `supervise_as` as Instance methods.

tarcieri:
Maybe make the breaking change on master and have 0-16-stable with API compat (for now)

digitalextremist:
Roger.

digitalextremist:
@tarcieri I am going to handle the breaking change in another issue. I want to be clear about which ones we're removing, because `supervise_as`, `supervise` and `pool` occur all over the place. I will ask some clarification questions in that new issue.

tarcieri:
Sounds good

tarcieri:
I think what might be particularly helpful is to extract supervisor configuration into a class separate from a supervisor-actor. People seem to want that sort of configuration object, but it would really help if it were a plain old Ruby object they can interact with in a single-threaded context

digitalextremist:
Sounds good. Will add that to the issue being written. For now, is this PR fine then?

tarcieri:
The duplication is pretty annoying, but c'est la vie...

Maybe wait for CI?

digitalextremist:
Agreed on duplication. And Travis is showing strange behavior. Will look into it.

digitalextremist:
:ship: The build errors are in the acceptable failures category, and in the tests marked flaky.

PR #507, opened 2015-03-25
pin the version of timers

punkle:
Fixes the incompatibility with versions of timers > 2.

In particular: ```/var/lib/jenkins/.rvm/gems/ruby-1.9.3-p484@global/gems/celluloid-0.14.1/lib/celluloid/receivers.rb:9:in `initialize': undefined method `new' for Timers:Module (NoMethodError)```

### DIFF
--- a/celluloid.gemspec
+++ b/celluloid.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |gem|
   gem.files        = Dir['README.md', 'lib/**/*', 'spec/support/**/*']
   gem.require_path = 'lib'
 
-  gem.add_runtime_dependency 'timers', '>= 1.0.0'
+  gem.add_runtime_dependency 'timers', '~> 2'
 
   gem.add_development_dependency 'rake'
   gem.add_development_dependency 'rspec'


tarcieri:
It should probably be `~> 4.0`

punkle:
Celluloid 0.14.1 uses Timers.new. Timers.new is not support in version 3 or 4 of timers.

tarcieri:
Oh, I missed what branch this was. Celluloid 0.14 is really out-of-date. Perhaps you should upgrade?

punkle:
It has been considered to upgrade. However there is quite an argument stacked against doing so. In my particular case Celluloid is a transitive dependency. The next version of the direct dependency is quite a leap and is not compatible. And that option would require quite an amount of work.

tarcieri:
I'd suggest just fixing this problem in your app's Gemfile then...

ioquatix:
here is a monkey patch

module Timers
  def self.new(*args); Group.new(*args); end
end

punkle:
@tarcieri @ioquatix thanks for the tips. :) A similar solution to the Gemfile pinning has been implemented and has fixed the problem for us (pinning using the gemspec).

Having said that I think this patchset solves the root cause of the issue. However, I do accept that this project has moved on many versions and cannot necessarily be expected to support an earlier version.

digitalextremist:
@punkle I had this issue too, and had to refactor my code to match the api changes. I feel your pain!

Issue #506, opened 2015-03-24
Remove supervision oriented instance methods. Implement Supervisor configuration object.


digitalextremist:
Per discussion on #503 and #505, we are making a breaking change to `master` but will keep the old-style API in `0-16-stable` ... the breakage will be as follows:

There will be no `supervise`, `supervise_as` or `pool` instance methods available actors.

All implementations of those methods will be `Class` methods going forward. No `Instance` methods available, as they are unsafe.

Additionally, supervision configuration will be handled by the creation of a new object which can be instantiated and prepared in a single-threaded context and potentially reused across various actors.

digitalextremist:
Here are the instances of `supervise', `supervise_as` and `pool` methods we currently have:

```
Celluloid::ClassMethods.supervise
Celluloid::ClassMethods.supervise_as
Celluloid::ClassMethods.pool

Celluloid::SupervisionGroup.supervise
Celluloid::SupervisionGroup.supervise_as
Celluloid::SupervisionGroup.pool

Celluloid::SupervisionGroup#supervise
Celluloid::SupervisionGroup#supervise_as
Celluloid::SupervisionGroup#pool

Celluloid::Supervisor.supervise
Celluloid::Supervisor.supervise_as

```

It seems the ones being annihilated are those instance methods on `SupervisionGroup`

And it seems the object being created is `SupervisorConfiguration`

/cc: @tarcieri

e2:
1. Please keep the master branch "releasable" at all times - you never know when you'll need a quick bugfix (and handling branches, backports, cherry-picking is all a big waste of time).
2. Don't break compatibility unless it's absolutely necessary to move forward - Ruby makes it easy to move code into deprecated classes which can be easily removed later (along with their specs)
3. At least keep the specs in sync with the changes - with a lack of docs, specs are the last bastion of sanity (`git blame` sucks as a replacement for documentation)
4. Give examples of the "configuration object" before you start implementing it - I don't see it as absolutely necessary (given the existing options are better documented)

A extra few comments:
- `Celluloid.pool` and `SupervisionGroup.pool` are unrelated (the first uses `PoolManager.new`, the second `PoolManager.new_link`)
- fix #508 should keep things backward-compatible and help clarify the argument handling (which isn't well defined to begin with)
- `Celluloid.supervise` is just a convenience wrapper for `Supervise.supervise` (which in turn is just a convenience wrapper for `SupervisionGroup.add`
[...]

PR #504, opened 2015-03-20
refactor per #503, fixing supervise and supervise_as handling of args

digitalextremist:
Fixes #503

PR #508, opened 2015-03-25
refactor out supervise* methods a bit + add tests

e2:
Refactor out `supervise` and `supervise_as` methods to external module to avoid duplication + add more tests for testing parameter handling (related to #506, #505, #504, #503).


digitalextremist:
@e2, thanks for the tests. Not sure if I'll merge this though because it seems to preempt #506. I will come back to this. There will be discussion about supervision I'm sure, I'll keep this branch in mind during that.

e2:
@digitalextremist - this is just further refactoring on 0.16-stable + some tests. Ideally the tests should be both for SupervisionHelper and as a shared example group in the other 2 places - but that's not worth the work if you're going to almost reimplement supervision from scratch, etc.

digitalextremist:
This is defunct as supervision is in its own gem now.

PR #513, opened 2015-03-28
shutdown quickly and gracefully when inside fork()

### LINKED ISSUE #468: Feature request: Celluloid could shutdown faster after fork()
I understand fork() renders Celluloid useless (in the child process).

I was just wondering if there was an elegant way for ActorSystem to detect that all the threads are dead, so mailboxes are dead, so there's no point in even calling actor.terminate! or actor.join.

(Found it while working on: https://github.com/rails/spring/pull/194)

How to preproduce:

```ruby
require 'celluloid'

Celluloid.logger.level = Logger::DEBUG
Celluloid.shutdown_timeout = 2

class Myactor
  include Celluloid
end

Myactor.new

Process.wait2 fork {}
```

e2:
Fixes #468

### DIFF
--- a/lib/celluloid/internal_pool.rb
+++ b/lib/celluloid/internal_pool.rb
@@ -6,6 +6,8 @@ module Celluloid
     attr_accessor :max_idle
 
     def initialize
+      @pid = $$
+
       @mutex = Mutex.new
       @idle_threads = []
       @all_threads  = []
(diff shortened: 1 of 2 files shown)


digitalextremist:
@e2 your branch/fork doesn't seem to show up -- where is this for me to merge it into stable? If it doesn't matter though, it could just go into `0.16.5` later.

PR #518, opened 2015-03-28
Silence $CELLULOID_DEBUG-related warnings

tarcieri:
For some reason explicitly initializing a global is causing test failures via spooky action at a distance. Kill me now.

@schmurfy this is your code that's breaking. Any ideas?

### DIFF
--- a/lib/celluloid.rb
+++ b/lib/celluloid.rb
@@ -3,6 +3,8 @@ require 'thread'
 require 'timeout'
 require 'set'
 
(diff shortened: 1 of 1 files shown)


e2:
Fixed in #519 - celluloid.rb was required after celluloid/rspec.rb, so it was set to false

schmurfy:
is this still an issue or do I need to have a look ?

tarcieri:
@schmurfy it's fixed

PR #519, opened 2015-03-28
Fix require order in specs so CELLULOID_DEBUG=true

e2:


e2:
Tests seem to pass, but the Travis job fails because of:

```
/home/travis/build/celluloid/celluloid/lib/celluloid.rb:31:in `actor_system': Celluloid is not yet started; use Celluloid.boot (Celluloid::Error)
	from /home/travis/build/celluloid/celluloid/lib/celluloid.rb:160:in `shutdown'
	from /home/travis/build/celluloid/celluloid/lib/celluloid.rb:149:in `block in register_shutdown'
```

e2:
Oh, gosh ... `$CELLULOID_TEST` is now false

tarcieri:
Thanks!

PR #522, opened 2015-03-31
Dead checking

digitalextremist:
Closes #350


digitalextremist:
The only failing tests now are `:flaky` ones, after the rspec syntax was changed.

Issue #359, opened 2013-11-26
please keep examples and changelog in the gem


boutil:
Hi,

Could you please keep in the gem the changelog CHANGES.md as well as the examples? It is useful to have them around. Moreover gems are now often the starting point for the packaging of Ruby libraries in Linux distribution (at least in Debian).
It would be nice to have them included in the package.

Thanks in advance.

Cédric

PR #523, opened 2015-03-31
Re-sync 0-16-stable with recent updates.

digitalextremist:
Bring over #518, #513, and #359 into `0-16-stable`


digitalextremist:
Everything important is green. Backporting...

Issue #334, opened 2013-09-20
Access to publish() through Celluloid class method.


digitalextremist:
Mentioned in IRC to @halorgium: IMO, a class method interface to publish messages ought to be accessible from outside actors, through a class method on `Celluloid`, just like `Futures` have such an interface.

For example, `Celluloid.publish()` or `Celluloid::Notifications.publish()`

digitalextremist:
@dilumnavanjana, this is referring to the `Celluloid::Future` concept, where instead of running `myActor.future.method()` we can run `Celluloid::Future.new { block }` ... for this issue, what I was asking about was the ability to avoid needing to run `myActor.publish()` ( `publish` within `myActor` ) and be able to use `Celluloid::Notificiations.publish()` directly -- a class method rather than an instance method.

For this issue, understanding `Futures` is not important, but to lightly explain those, you can run a piece of code in a `Future` and it will happen in its own thread, asynchronously, then when it's done you can obtain the output by running `.value` on that future. I will give you a link to that separately, to keep this issue focused on the `.publish` class method.

dilumnavanjana:
Is this ```.publish``` class method something similar to,

```ruby
def self.new(*args, &block)
      return super unless block

      future = new
      Celluloid::ThreadHandle.new(Celluloid.actor_system, :future) do
        begin
          call = SyncCall.new(future, :call, args)
          call.dispatch(block)
        rescue
          # Exceptions in blocks will get raised when the value is retrieved
        end
      end
      future
    end
```
in ```Celluloid::Future```?

digitalextremist:
Yes, in the sense that is a class/module method. Are you familiar with class/module methods? Maybe we should do this by email to keep this ticket clear. What we're looking to do is implement a `publish` class method, on the `Notifications` module

 It seems pretty straight-forward. If you look at the `#publish` method here:

https://github.com/celluloid/celluloid/blob/master/lib/celluloid/notifications.rb#L7

I would need to test this ( or this would be a good experiment for you to try ) ... but this one line below seems like the solution, now that I glance at the above code in my link:

```
   module_function :publish
```

That would treat the instance method `#publish` as also being the class/module method `.publish` ... which would be all that's needed since the method itself just relays a notification to the `Fanout` actor.

It'd be different if it were a `.subscribe` which would need much more work, and would probably be a bad idea to directly access as a class/module method.

If I make that change, would you test it?

PR #527, opened 2015-03-31
Publish module method, and pub/sub example.

digitalextremist:
This not only adds support to cover #334, it also brings in an example for the wiki so we don't need to only reference `Reel`

### DIFF
--- a/lib/celluloid/notifications.rb
+++ b/lib/celluloid/notifications.rb
@@ -82,7 +82,7 @@ module Celluloid
     end
   end
 
-  def self.publish *args
-    Notifications.publish *args
+  def self.publish(*args)
+    Notifications.publish(*args)
   end
 end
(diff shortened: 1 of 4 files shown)


digitalextremist:
@tarcieri for future reference, as with here -- is it more efficient to create a method which redirects to another method as with 2d7a982, or is it better to `extend Forwardable` and delegate one method. I always think extending `Forwardable` for one method isn't worth the trouble if it's going to impact so many other things downstream in something as large as `Celluloid`.

lib/celluloid/notifications.rb
```diff
@@ -80,4 +81,8 @@ def matches?(subscriber_or_pattern)
       end
     end
   end
+
+  def self.publish *args
```
tarcieri: Parens?

I guess we should get Rubocop hooked up for real...

tarcieri:
This sort of thunk is fine if it's just a method

lib/celluloid/notifications.rb
digitalextremist: fixed in 15d5cfe

PR #532, opened 2015-04-01
Separate testing using fibers vs threads

e2:
Benefits (development):
- task based tests aren't run twice in dev
- ~~shared examples can now be refactored out~~
- slightly less tests during dev before PRs
- better insights in case of Travis failures
- extra test covered in stack dump

Downside:
- larger travis build matrix (can be selectively reduced as needed)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 script: rake ci
+language: ruby
 rvm:
   - 1.9.3
   - 2.0.0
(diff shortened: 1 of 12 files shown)


e2:
@tarcieri - would it be ok to undo the "shared examples" thing because of this (and have Celluloid::IO specs include the specs directly via gem path expansion instead of requiring shared examples)?

tarcieri:
Sure, sounds good to me

digitalextremist:
@e2 I'm going to review the *many* contributions you've made today by merging this branch and the previous one into an overall test refactoring branch, hopefully with @niamster and/or @asmod4n perhaps bringing more eyes. So these moved over to #533 & #534. Any more changes you have to the test-suite, please call for pulls into `refactor-test-suite` so we can handle review in one large sweep.

e2:
@digitalextremist - ok

Issue #526, opened 2015-03-31
Test overhaul.


digitalextremist:
This involves the following, but is broader than these:
- [x] #524 Deal with `:flaky` tests.
- [x] #525 Get travis working on `jruby-head`
- [x] #528 Radical amount of warnings in specs [ @e2 in #530 ]
- [x] Eradicate timeouts on all RVMs.

In general, we need to refactor how tests are handled, removing intermittence; not disabling tests but removing the flaky factor -- which seems largely about putting in some strategically placed `sleep` calls and removing reuse of threads and fibers.

While we don't need to go so far as to remove `allowed failures` per se, it would be nice to be able to detect defects in bleeding-edge RVM's with Celluloid, rather than coming to expect failures.

digitalextremist:
@e2 are you willing to team up on this?

e2:
I can't make any promises. I think I'll start a new Celluloid branch, because debugging tougher issues may need lots of codebase changes (and so, PRs won't make sense) - and I'll "backport" changes to `master` on a case-by-case basis. I'll probably need to IRC for this quite a bit...

I recommend merging in any pending PRs first - because I'll likely have to reorganize/customize things in the codebase and logic to debug/work/fix faster.

e2:
@tarcieri - what's the "official status" behind ThreadTask vs FiberTask? I have the impression that ThreadTask provides backtraces, but is slower/more flaky - while FiberTask over works well, though doesn't provide as much useful debugging info (and I don't know about JRuby). Not to mention that Fibers are really threads on JRuby anyway (along with all the other consequences).

digitalextremist:
Sounds good. I'm trying to clear out at least everything in the `0.16.5` milestone, and I'm picking off several `0.17.0` issues also. By the way... this `Probe on boot should capture system actor spawn` failure happening in `2.0.0` is a carry-over from `$CELLULOID_MONITORING`

digitalextremist:
@e2 there is a patch out providing backtraces on `TaskFiber` -- I'm thinking we should really merge that.

digitalextremist:
I think this `Probe on boot should capture system actor spawn` failure is `:flaky` but I'm getting mixed results.

e2:
@digitalextremist - probably worth checking out/cleaning up the project branches too. future-cancel sounds cool...

tarcieri:
@e2 both are supported. TaskFiber is definitely more stable. @mperham was trying out TaskThreads in Sidekiq.

[...]

Issue #524, opened 2015-03-31
Deal with :flaky tests.


digitalextremist:
Tests marked `:flaky` either need to be fixed, or removed from the test suite ( commented out to be addressed later ).

e2:
Failing tests marked as `:flaky` should be currently be rerunning - both locally and on Travis, so I see no reason to remove them, really.

I chose these values: https://github.com/celluloid/celluloid/blob/master/spec/spec_helper.rb#L57 because the tests should work "most of the time" and fail only "occasionally". If a test fails "more often" it's better to put a small sleep() here an there (as much as I see them as pure code smells) and get it "working most of the time" again.

Flaky tests are actually a side-effect of not using isolated tests (e.g. with mocking). Ideally, Celluloid itself shouldn't be using Threads (or Fibers on JRuby) at all.

I'm just saying that the "flakyness" of tests is a symptom of much bigger problems.

Other than that - Celluloid needs an overhaul on the "testing support" side. I'm thinking of things like CELLULOID_DEBUG, CELLULOID_MONITORING, mocking task classes, Travis environment variables for task class (and refactor out the RSpec shared examples).

It's too easy to scapegoat "flaky tests" - and disregard actual, genuine race conditions in Celluloid (or even bugs in  JRuby).

I think each "flaky" test should have it's own issue (or grouped at least if they are related).

e2:
Some tests just need to marked as flaky, e.g. if this build today failed because this test if flaky - then this test simply needs to be marked as such (it isn't as of master):
https://travis-ci.org/celluloid/celluloid/jobs/56532489#L643

digitalextremist:
I wholeheartedly agree @e2. I'd be interested in cleaning these up with you if you're willing.

digitalextremist:
In dealing with `:flaky` tests, it seems one issue is killing Threads between tests, which under jRuby means killing Fibers also, which causes a melt down.

But then when you avoid the meltdown, the tests start failing intermittently, because now there are undead Threads/Fibers wandering around.

digitalextremist:
A recent search through the codebase(s) shows no `:flaky` tests. It seems we can remove the evidence of there ever having been, then this is done.

e2:
@digitalextremist - this can be closed. If there are flaky tests, those should be individually opened as new issues.

digitalextremist:
We just need the metadata checks and environment variables allowing then removed.

e2:
@digitalextremist - already done on my experimental branch

PR #535, opened 2015-04-01
Refactor test suite.

digitalextremist:
This is a single review point for @e2's many changes to the test suite per #526, and any other contributions/revisions brought in. Please address revision PR's to this branch `refactor-test-suite`


digitalextremist:
Now celluloid/celluloid-io#134 is depending on this code.

digitalextremist:
@e2 what's your reasoning for marking `1.9.3` as an allowed failure?

digitalextremist:
IMO, the stable versions of `jruby`, `rbx`, and `1.9.3` ought to be mandatory successes, we have a *lot* of code deployed in those environments. Where are you coming from with that thought @e2?

e2:
@digitalextremist - ironically, the failures are probably just test-env specific. 1.9.3 can be mandatory already, since it's green. JRuby may take time to get to green, and Rbx might be a lost cause (not enough support to deal with the complexity of the test environment + threads).

It's not hard for a test env to crash JRuby or RBX - especially because RSpec does a lot of magic there. Also, the Celluloid boot/shutdown between each tests isn't "normal".

What I'm saying is: it may not even be possible to get the all the tests green by "fixing" them - it may become just about side-stepping bugs in interpreters (and not Celluloid itself).

digitalextremist:
@e2, OK. I hear you. For now I'll settle for `1.9.3` being mandatory out-of-the-box, and we can think further about beating the high performance twin RVMs into submission somehow.

digitalextremist:
*mandatory

digitalextremist:
@tarcieri the CI tests doubled ( again ). If this is a problem I will refactor that, but I set these conditions, following @e2's changes. This will only last until #524 is resolved.

* Use `TaskFiber`, bypass `:flaky`
* Use `TaskFiber`, do not bypass `:flaky`
* Use `TaskThread`, bypass `:flaky`
* Use `TaskThread`, do not bypass `:flaky`

Then all but use `TaskFiber`, bypass `:flaky` are allowed failures.

digitalextremist:
I wonder if we should just not use `TaskThread` in CI and expect people to test their own situation themselves, by setting `CLLLD_TASK_CLASS=TaskThread`

e2:
@digitalextremist - just hang on for now. I've almost got all TaskThread tests working on JRuby head. Many flaky tests are no longer flaky - and most fixes are due to race conditions and internal termination handling bugs/problems in Celluloid. Just hang in there...

@tarcieri - I'll have a branch put up once I do cleaning up. I'd like to ask for a thorough review though (since I'm making LOTS of assumptions about Celluloid that could be completely wrong or result in horrible performance). Some changes are drastic, like completely reworking how Actor.join is done, disabling termination timeout in favor of specific timeouts per Actor, a potentially slow (for now), mutex-heavy InternalPool (but one that works almost flawlessly in tests, no matter how much banging the tests get), avoiding Thread.kill completely (since you'd need to Einstein to use it correctly on JRuby), etc. Just giving a heads up. ETA - a few days (still some StackDump, Probing issues and supervision issues).

e2:
@tarcieri - I disabled some tests that use Actor.join directly, because this method/API doesn't make sense to me. (If at all, people should just shutdown the Actor instead - it just potentially hangs an app or tests if no timeout is given, while a timeout just destroys any state consistency on JRuby). ThreadHandle.join doesn't make sense to me either (doesn't seem to be used anywhere, either).

tarcieri:
@e2 I can have a look when you're done.

I'm glad you're working on InternalPool as the current implementation was completely broken, but I think I'd actually prefer ditching it completely. It's been the source of a lot of memory leaks, and I think using a more normal thread lifecycle would help avoid those. Just as an example it clears Fiber-local `Thread#[]` but doesn't clear "real thread locals" i.e. `Thread#thread_variable_get/set`, nor does it ensure the death of all Fibers for a thread. Given the complexity involved, I think ditching InternalPool completely is the best option.

digitalextremist:
@tarcieri, yes I'm ditching it currently  ( fyi, @e2 )

e2:
> I'm glad you're working on InternalPool as the current implementation was completely broken

Simply using arrays without locks or instead of queues on JRuby mean "broken".

> but I think I'd actually prefer ditching it completely.

For production - I agree 100%. Even if it's "reimplemented". Though, I'd leave the "fixed" version in spec/support for the sake of "stress testing", especially for JRuby and Rbx.

> It's been the source of a lot of memory leaks, and I think using a more normal thread lifecycle would help avoid those.

It would make debugging much easier if there was just one-thread per task. Not to mention the actual busy/idle management was not clearly defined. Or tested.

> Nor does it ensure the death of all Fibers for a thread.

That actually bit me, because one of the stack_dump spec uses a Fiber and Thread actors. I'm trying to think of threads as coroutines, because they are used that way in Celluloid. Especially to have the same interface as Fibers for task_class.

In fact I learned, Fibers actually can be killed individually on JRuby, because they are threads visible in Thread.list: https://github.com/jruby/jruby/issues/2773

I haven't got to the specs running in Fiber mode yet - I want to make sure all tests pass with Threads first.

> Given the complexity involved, I think ditching InternalPool completely is the best option.

Even if tests are green after retries on Travis ... it doesn't mean the race conditions aren't bugs in Celluloid. So having a "fixed" "reimplementation" of InternalPool could be useful as a helper in spec/support.

It's too easy to "accidentally" get tests green, when they shouldn't be green. Especially since it's hard to predict the side-effects of almost any PR at the moment.

I'd rename the class (so people know not to use it) and have Travis test on the fixed class anyway - until JRuby and Rbx support are good enough. Because Celluloid doesn't really have "unit tests" - they're actually all non-deterministic "stress tests", so even bad implementations have a lot of benefits right now.

What do you guys think?

digitalextremist:
Thanks again for your attention to this layer of things @e2. I think we should allow `InternalPool` to be replaced the same way we do with `.task_class` for Tasks. We can use `.internal_threads_class` or something like that. We can default to the new `Class` coming in for now, then at some point we can totally remove which ever route is least performant and stable... or just leave it there, usually unused like `TaskThread` is. The way `Task*` all "behave as" `TaskThread` is how I'm doing the `InternalPool` replacement -- the API is the same for both... with certain special distinctions which are non-critical differences.

digitalextremist:
I will move `InternalPool` over as part of what I'm doing, unless there's objection to the approach I gave just now. This will allow us to switch on and off the internal thread "collection" just like we do task handlers right now.

digitalextremist:
`Celluloid.group_class` seems short, sweet, and unambiguous.

e2:
> I think we should allow InternalPool to be replaced the same way we do with .task_class for Tasks. 

We think alike - I actually wanted to mention that above - along with the idea to reduce the API:

Users may "want" things like `busy_size` and `idle_size` or have info about threads for the `StackDumper` - but instead, really, it's all available in `Thread.list`. 

(It should be enough to add some examples of things like `Thread#celluloid?` and wrapper class usage to match threads with actors, etc.)

Yet, the biggest reason to drop this extraneous API is: it complicates the internals. 

I've considered e.g. having `busy?` (instead of a useless count) - but anyone tweaking parameters such as max_threads_for_something needs to base calculations on the status of all threads used in the app.

I said useless, because of e.g. "race conditions" for counting existing threads (e.g. a thread can be created and busy - and not yet make it into the internal array). And recycling of threads for tasks makes that exposed info too pool-implementation specific.

As for tests, things like `busy_size` can be replaced with proper mocking (checking e.g. for the right number of get_thread invocations).

digitalextremist:
I agree across the board with what you said. To me that translates to creating a base class the implementations all start from ( like `Reel::Server` ) and we `raise NotImplemented` or something like that for unimplemented/undefined methods... if we even need to do that to keep the tests accurate. Otherwise I am well on my way with `Celluloid.group_class` and sort of wishing `refactor-test-suite` were merged into `master` already, but it'll be fine. I don't think I'm really making conflicts.

e2:
> if we even need to do that to keep the tests accurate

There's too much second-guessing as a contributor. Like, I have no idea if `busy_size` is a "major feature" ... or just there for the tests. (I guess git blame could answer it somewhat).

I think the basic API would be `:get, :put, :grow, :shrink, :shutdown`. grow/shrink = by one - and is only there to "allocate" system resources (so e.g. you can guarantee the pool won't need to allocate new threads after starting - which could potentially fail).

digitalextremist:
@e2 for now let's have the standard be `initialize`, `get`, `shutdown` ( and `kill` ), `create`, `each` ( and `to_a` ) **and that's it** ( pending me forgetting something obvious ) ... Things like `grow`, `shrink`, and `put` are specific to one type of `Group` or more, not all. This non-pool `Group` doesn't need those, so those and `busy_size` and `idle_size` ( which are more for the community to use than `Celluloid` itself ) are not going to be over-ridden base methods.

digitalextremist:
and `running?`

e2:
`each` and `to_a` are either racy or troublesome with mutexes (they can change the flow of things quite a bit - expecially if you query at the wrong time). Using arrays without mutexes (specifically, reading from arrays) on JRuby is a big no-no. They also prevent you from using queues, which don't have the overhead of locks.

digitalextremist:
I will implement this new type of `Group` with `each` and not `to_a` then we can look at its raciness. I'm using `Queues` ... it might not be racy the way I'm doing it.

e2:
As for `running?` - that isn't necessary. Conceptually, it's a "pool" - and you are only about taking or freeing resources, nothing else. `grow` and `shrink` are just optimizations/safety measures. It also raises questions, e.g. is it still `running?` just after calling `shutdown?`. Technically, if you're not just killing everything indiscriminately, it's still "running" until shutdown finishes. You could define 'running' as "empty queues" - but then you have the race condition, because you can allocate and start a thread - without being able to register it (the thread can start running before you have a chance to add it to an internal queue or array).

digitalextremist:
I agree, `running?` and `active?` will be `Group::Pool` specific... which involves an `ActorSystem` tweak I think, but that's fine.

e2:
I got rid of 'kill' actually (had to rework the actor system shutdown code myself). The biggest problems were when shutdown hit the timeout, and then kill was invoked. In JRuby "anything can happen at any time". Literally.

tarcieri:
`Celluloid::Pool` is mostly just broken and needs to be replaced by something else ala `celluloid-coordinator`

digitalextremist:
So far it *seems* like `Celluloid` runs a lot faster without `InternalPool`

e2:
:+1: for a "Celluloid-based supervised pool"

e2:
PoolManager + InternalPool = horrible experience

digitalextremist:
With how radically `TaskPooledFiber` helped things, I'm really curious to see `PoolManager` yanked, as well as `InternalPool` ... and yeah, with those both in the same sentence ... `kill -9 self`

digitalextremist:
I am starting to wonder if removing `InternalPool` and putting in this change I'm about to push isn't going to somehow unmaim `PoolManager` and how we'd know if it did.

digitalextremist:
This is no longer valid, but I'm hesitant to close it in case @e2 is going to hit a wall without it. In a moment though I will be closing this and opening a new PR merging `0-16-5-prerelease` into `master`

digitalextremist:
This is getting convoluted, sorry @e2 I'm switching destination branches to the one linked above:

## The entire test suite passes in under 30 seconds without `InternalPool`*

Except for some lingering `:flaky` tests that sometimes make it ~45-65 seconds.

digitalextremist:
Moved ( for the last time ) to #547

PR #537, opened 2015-04-01
Rework shared examples

e2:
Fixes on top of #534:
- workaround to prevent 1 spec from crashing JRuby (JRuby lists internal Fiber thread in Ruby's `Thread.list`)
- Ruby < 2.x fixes for getting Task class from ENV
- reorganize shared examples, so that at least actor specs aren't in shared examples (makes dev harder)

Celluloid::IO specs will need to also updated to work with these changes (comming up...).


digitalextremist:
@e2 thanks for pushing to this branch. I'll keep snowballing #535 with all your contributions, then we can revisit celluloid/celluloid-io#134 once that lands.

tarcieri:
At some point I'd like to get most of the RuboCop violations `--auto-correct`ed and have a sane RuboCop policy too.

e2:
@tarcieri - yes, that's probably best done after most PRs have been pulled in and there aren't many feature branches (otherwise it can result in a ton of unnecessary merges/conflicts).

e2:
I like how you can immediately see that the hanging tests are due to failing TaskThread tests (JRuby): https://travis-ci.org/celluloid/celluloid/builds/56668865

digitalextremist:
@e2 that is a *substantial* improvement.

Issue #538, opened 2015-04-01
Set RuboCop loose in codebase.


digitalextremist:
> At some point I'd like to get most of the RuboCop violations --auto-corrected and have a sane RuboCop policy too. -- @tarcieri ( #537 )

After `0.17.0` or `0.16.5` is ready for a prerelease, and all the #529 branches are audited, implement Robocop policies and bring the codebase up to standard.

e2:
I've "hidden" the current offenses here: https://github.com/celluloid/celluloid/pull/539

It's best to decide which ones are essential to reenable (by simply removing them from the `.rubocop_todo.yml`, and which ones to ignore for now (like line-length).

Probably the best to start with are the infrequent ones that can be autocorrected - as long as it doesn't interfere with merging, etc.

/cc @tarcieri

tarcieri:
I'd suggest blowing away any extant config, running --auto-correct, then getting a baseline.

I am guessing we'll have to crank up things like ClassLength/MethodLength/AbcComplexity up quite a bit.

Beyond that, we can bikeshed about various details. I'm a fan of `Style/StringLiterals: EnforcedStyle: double_quotes` :wink:

e2:
@tarcieri - I did that. The `.rubocop_todo.yml` in #539  is a "fresh list" of all the offenses.

> I am guessing we'll have to crank up things like ClassLength/MethodLength/AbcComplexity up quite a bit.

RuboCop already took care of that - it displays the largest found size (176 characters - if someone creates a longer line, it will be visible, while anything less - will be skipped).

> I'm a fan of Style/StringLiterals: EnforcedStyle: double_quotes ￼

413 offense - I wouldn't do it because merging could get in the way.

I did a test - here's what auto-correcting ALL the current offenses would look like:

https://github.com/e2/celluloid/commit/c6d6e1d45cd3f83f4afac0a766154acd7c647e68

No major benefit really in the current codebase IMHO.

e2:
I just switched to double quotes (less offenses): https://github.com/e2/celluloid/commit/64faa823e73f6d408aacb9afc8f399ccdf70dad0

digitalextremist:
@e2, you are a monster.

I agree, this is largely trivial at the moment and best run at the turn of `0.17.0` when we're within range of `1.0` ... otherwise this is just going to obfuscate our remaining steps to release. But it's all set though.

tarcieri:
[...]

PR #539, opened 2015-04-01
Hide RuboCop offenses #538

e2:
Disabled all the current RuboCop offenses using the .rubocop_todo.yml.

(Offenses listed there can be selectively re-enabled and fixed on an ongoing basis).


digitalextremist:
@e2 what about keeping the live file and just commenting them all out for now?

digitalextremist:
@e2 do we need this? Is this going to matter long enough to commit it?

e2:
Yes, but let me know when you're done with all work and ready to merge to master (it should be the last step given all the work you're doing) and I'll re-do this commit (to reflect the codebase).

After that, you and Tony should go through the list to decide what to enforce and what should stay as "ignored". And *then* when can close this, and *then* someone can fix the offenses so you and Tony and happy.

Otherwise, the RuboCop version bump alone is important.

digitalextremist:
@e2 roger that; bump the version for now, in `0.17.0-prerelease`, then we'll come back to this and I will pair with @tarcieri to comb through this before we push `0.17.0`

digitalextremist:
( @tarcieri, I was hoping to ship `0.17.0` before GSoC started officially. But it might help to go through `0.17.0` in person, for stuff like this, and to get first hand from you the remaining roadmap versus what right now I've projected. I need to go to Mexico in the first week of May, but then after that I can probably come up for end-of-release stuff like this; back-of-the-mind FYI )

digitalextremist:
As it stands, everything in the prerelease is currently marked "false" without inheriting from a secondary file, so I'm going to close this and we can readdress it at #538 later.

PR #545, opened 2015-04-05
0.16.5 Prerelease Track

digitalextremist:
Work in progress. Obsoletes #535. Closes #520.

`InternalPool` is disabled ( and broken ) at this point, but `Group::Proactor` is insanely fast and working. Going to start hitting this with `TravisCI` and get `InternalPool` back, and hopefully catch @e2's changes here. This will now be the branch that will become `0.16.5` ( and not be merged into `0.16.0` stable ) since we're closer to completing the next milestone than focusing on backporting to the last.

Issue #520, opened 2015-03-29
Research removal of InternalPool completely.


digitalextremist:
Per Garbage Collection observations made in #463 -- research the ramifications of removing `InternalPool` entirely. If the refactor can be performed safely and in reasonably good time, benchmark the performance of Celluloid without `InternalPool` and with it, and then retest #463 without it.

Of course, researching this truly and completely will result in an experimental prototype of Celluloid with no `InternalPool` ... if that branch outperforms the existing form *and solves the Garbage Collection dilemmas we're experiencing* polish that experimental version and remove `InternalPool` going forward.

digitalextremist:
See comments on #535 regarding `.group_class` ... about to push first run at this, working fine, all tests running stupid fast, but I need to retroactively heal `InternalPool` ( or not )

digitalextremist:
Closed in #545

Issue #511, opened 2015-03-28
Refactor supervision.


digitalextremist:
This started in #506 but for clarity here is a clean ticket. This is a big breaking change.

- [x] Remove all instance methods for supervision.
- [x] Implement a configuration object for actor supervision.
- [x] Add back support for supervision trees ( basics, complete this in `0.17.5` )
- [x] Fix regressions in `Pool`

Also clean up the `supervision` label:

~~#493 null pointer on SupervisionGroup#terminate~~
~~#434 Calling async.run from base class constructor might not be a good design.~~
- [x] celluloid/celluloid-supervision#3 ~~#430~~ supervisors_and_registry.rb not working correctly
- [x] celluloid/celluloid-supervision#4 ~~#355~~ Support for dynamically removing SupervisionGroup members
- [ ] celluloid/celluloid-supervision#5 ~~#331~~ Supervisors attempt to restart the same actor if terminated before restart completes
- [ ] celluloid/celluloid-supervision#6 ~~#233~~ Exception in #initialize took down parent SupervisionGroup

e2:
:+1

digitalextremist:
Refactor as gem, per #562.

Senjai:
@digitalextremist :+1: Sorry I wasn't able to follow through on the original issue, work caught up with me. I will keep watch though! I dont know why there's so much talk about extracting things into seperate gems. I don't know where the benefit is unless they can be run on their own. I'm all for removing as much knowledge from the system as possible though.

tarcieri:
@Senjai moving thing out of core allows us to focus more on the stability of core, while accepting and informing people of different levels of stability for the gems.

It also allows us to stabilize the API of core (for e.g. a Celluloid 1.0 release) while leaving the unstable parts to mature separately in gems.

Senjai:
@tarcieri that makes sense, so they would take more of a plug-in role?

tarcieri:
Yes. It also allows for parallel development of replacements (e.g. `celluloid-coordinator` should replace `celluloid-pool`). If these features are gems and not core functionality, I think people might be more interested in evaluating alternatives.

digitalextremist:
The sub-issues in this are all going to be moved to the issue tracker in their own gem and closed here, but this will remain for overall status.

digitalextremist:
Now, we all know the benchmarks on CI are prone to outside fluctuation and mean basically nothing for performance testing, only emergency detection. But..

### `ruby-2.2.0`

[...]

PR #547, opened 2015-04-05
0.17.0 Release Track

digitalextremist:
After https://groups.google.com/d/msg/celluloid-ruby/o1sm7gkMCsQ/iAVBP6EX4T4J
`0.16.5` is being bypassed as the release is far larger than previously planned, faster than planned.

#545 obsoleted, #520 closed, #535 will be completed here, #544 solved already... etc.
Closing in on a 10.17.01 release... abuse of the code is invited, and any feedback/fixes of course.


digitalextremist:
I've got this deployed on my development server, using a complex instance of `Celluloid`. And while the tests all pass, something isn't right. Digging into it, but we've gotten a lot moved forward today.

e2:
@digitalextremist - update: no ETA yet, because I don't consider things "working" until all JRuby tests pass (testing on MRI only simply isn't testing at all).

I've changed my opinion - I'd say that if only one Ruby had to be mandatory, I'd make it JRuby (and not MRI). E.g. the reason JRuby "hangs" (in some places) is because you simply don't get a deadlock exception on JRuby (because a deadlock can only be certain on MRI due to the GIL). And "real deadlocks" are more likely on JRuby (so if something works on MRI, but hangs on JRuby - it's a deadlock bug in Celluloid, and not something wrong with JRuby).

There are major changes in the way things flow, so I won't start merging until JRuby is green. I'll branch off and rebase whatever branch you want, since I'll be cherry-picking things commit-by-commit anyway.

Don't hold up because of me (since I have no ETA yet - still trying to get tests passing). I don't see any major merge issues here anyway (I branched off of refactor-test-suite to begin with). 

Whatever changes you make here, all I care about is all 0-16 tests either passing on MRI - or simply "disabled" (as opposed to "removed"). If you need to remove tests or methods - it's best to mark them as such (since I've reorganized some tests and made a few useless ones useful again).

Just saying in case I'm stuck for longer.

lib/celluloid/groups/proactor.rb
```diff
@@ -0,0 +1,56 @@
+require 'thread'
+
+module Celluloid
+  class Group
+    class Proactor < Group
```
tarcieri: Why this name? Proactor has a very specific meaning in asynchronous I/O

lib/celluloid/tasks/threaded.rb
```diff
@@ -0,0 +1,55 @@
+module Celluloid
+  class Task
+    # Tasks with a Thread backend
+    class Threaded < Task
```
tarcieri: Why are you changing these names?

digitalextremist: For consistency and following precedent. There is `Task::` and now `Group::` which are both ( like `Server::` were, moving from one to two to many ) in their own directories.

I hesitated to change the directory names to singular. I think plural may be better, but wasn't sure of changing `server/` to match precedent at every point.

Both tasks and groups are organized by type within one namespace rather than diluting the general namespace. And we have `PooledFiber` dropping into `Task` as well, but in its own gem. Something about a gem adding to the general namespace to add a specific type of something we have several of felt like we officially and markedly went over the point where there is one type of task, one internal pool and no other way around it. Now we encourage overriding those very loudly.

lib/celluloid/groups/proactor.rb
digitalextremist: Yeah, I'm trying for that pattern. The implementation isn't finished yet but there's a finalizer that's not a preset. Actually my default needs to be run in every case but another finalizer can be prepended. Not insisting on the name but am trying to correctly implement this pattern  ( not exclusively as I/O however ).

This gets instantiated then it makes and kills threads, so calling it itself `Thread` seemed wrong, so I'm focusing on its behavior.

tarcieri: I don't get how "creating a new thread" relates meaningfully to the Proactor Pattern

digitalextremist: Well there is thread creation, but it's sort of a high integrity thing going on here:

0. Create thread.
0. Small to large  ( possibly blocking ) activity.
0. Conclude, blocking.

I focus on 2 & 3 more, and how that's handled, trying to fit a pattern, otherwise I'd just call it Dispatcher and call it good.

tarcieri: I'd probably call it something that describes what it's actually doing like `Spawner`

digitalextremist: I'm down with `Spawn` or `Spawner` ... and I'll just hard code the finalizer as purely clearing locals and abandon the pattern I was shooting for, for now. I could see a use for different `Group` classes, not only for `ActorSystem`

digitalextremist:
Wow, 2e0bbe6 was a critical issue, and my "catch all" overly-complex `Celluloid` implementation caught the case where a `Hash` being passed into `supervise_as` was interpreted as an options `Hash` and not sent along as an argument. This was the "not quite right" thing I was picking up on.

digitalextremist:
Will mop up errors popping up related to `SupervisionGroup` shortly; may just launch into #511 instead.

digitalextremist:
@e2 we now have another test that cannot be passed by `Spawner` -- I'm updating it to *not* keep sleeping threads, since it's not a pool. So we need a way to flag which tests conform to `Group` standard and which are special. I think `Group::Pool` is going to need additional tests to be outside the `it_behaves_like` so we can revert temporary methods and let `Spawner` be leaner.

digitalextremist:
@e2 but the amendments you are making are overall *very* awesome.

digitalextremist:
@e2 `Group::*.to_a` can stay, but it's going down to the superclass.

digitalextremist:
Same with `each`

e2:
@digitalextremist - I'd aim for a leaner Spawner and Pool no sooner than JRuby works - some of the stuff is just essential for debugging when things go wrong. And too much is going wrong right now. 

Not to mention - some races are reproduceable only with "extra stuff" happening. Heisenbugs need to be squashed before "refactoring".

Otherwise, `:each` should be sufficient as a enumerator. Though, I'd wonder why someone even needs an enumerator. I'd use a queue-based "event listening" (threads created, threads finished, etc.) - which could happen without any locks, anyway.

digitalextremist:
`StackDump` uses an enumerator, and so will `Group::Unlocker` but don't worry about thinning `Spawner` or `Pool` for now. Are you sure they're not Schrödinbugs? :)

e2:
> Are you sure they're not Schrödinbugs? :)

I wouldn't be surprised if some things worked simply because of JRuby bugs. (Once/if they'd get fixed, we'd uncover Schrödinbugs). Not saying JRuby is bad - just that a lot of stuff that's happening in celluloid isn't covered by JRuby tests.

digitalextremist:
@e2 beware, something very large just broke, you may want to go get some coffee.

digitalextremist:
@e2 problem solved, a downstream `each` definition was washing out my new superclass method.

digitalextremist:
I think it's time to go get coffee before I accidentally destroy the universe.

digitalextremist:
@e2 we can leave this open, no longer bombard people with notifications for a while, and both write to abstractive/celluloid#4 onward. Once done there, we will merge abstractive/master into this branch again.

You have an invitation to write into that repository, and I will add new remotes to my local repository, or just start using a separate clone.

digitalextremist:
This is now the main focus of attention for `0.17.5` again. It's wildly out of focus but it'll be cleaned up over the next few days, then issue by issue, pr by pr concentration will resume clearly.

digitalextremist:
@tarcieri, @niamster ... this needs cleaning and probably some polishing up, which I will still do. But it is not only running without problems so far -- completely refactored as promised -- but it is also backported #614 and compatible with 0.16.0 expectant code. Whatever's left from this is to remove lingering `#de` comments, maybe some more consolidation of code used across gems, possibly a little bit of sanity checking, etc. We're basically there. @dilumnavanjana this would be good to thoroughly review.

PR #552, opened 2015-04-06
Implement Future#cancel

digitalextremist:


tarcieri:
This leaves the future running and silently swallows the response?

e2:
@tarcieri - yes, I believe the idea is simply to "mark" some futures in an array as canceled (due to conditions or timeouts) - while still continuing to iterate and wait on other futures, e.g.

```ruby
futures = workers.map { |w| w.future.foo }
# ( some time goes by ...)
futures.map(&:cancel) # cancels slow workers, who haven't finished yet
finished_values = futures.map { |f|  f.value rescue :canceled }
```

PR #553, opened 2015-04-08
Test suite fixes

e2:
Mostly test suite related fixed/patches.

Since threads are completely dysfunctional on this branch, it might be worth dropping support for Threads completely (the only other option is to get Threads to behave *exactly* like fibers - which is probably not worth the time and maintenance).


tarcieri:
One note since I don't have time to do a more in-depth review:

1.9.3 is EOL. We should probably test against 2.0.0 instead.

digitalextremist:
This is merged into the prerelease branch so we can keep working on it there. Looks exciting. We can review the prerelease as a whole, later.

e2:
Thanks! I'll need to rebase now...

PR #555, opened 2015-04-08
use instance_eval to fix injections

e2:
Not enough tests, and I'd use Rails-like before/after filters instead ... but this is needed for: https://github.com/celluloid/celluloid-pool/pull/2

### DIFF
--- a/lib/celluloid/supervision_group.rb
+++ b/lib/celluloid/supervision_group.rb
@@ -145,20 +145,27 @@ module Celluloid
         @args = prepare_args(@options['args'])
         @method = @options['method'] || 'new_link'
 
-        @injections[:initialize].call if @injections[:initialize].is_a? Proc
+        # TODO: rename to ":after_initialize"?
+        invoke_injection(:initialize)
 
         start
       end
       attr_reader :name, :actor
 
       def start
-        @injections[:start].call if @injections[:start].is_a? Proc
+        # TODO: rename to ":before_start"?
+        # TODO: not tested
+        invoke_injection(:start)
+
         @actor = @klass.send(@method, *@args, &@block)
         @registry[@name] = @actor if @name
       end
 
       def restart
-        @injections[:restart].call if @injections[:restart].is_a? Proc
+        # TODO: rename to ":before_restart"?
+        # TODO: not tested
+        invoke_injection(:restart)
+
         @actor = nil
         cleanup
         start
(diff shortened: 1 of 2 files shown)


lib/celluloid/supervision_group.rb
```diff
@@ -176,6 +183,11 @@ def cleanup
 
       private
 
+      def invoke_injection(name)
```
digitalextremist: Extremely good eye, and approach here @e2.

spec/celluloid/supervision_group_spec.rb
```diff
@@ -58,5 +62,19 @@ class MyGroup < Celluloid::SupervisionGroup
...
+      group_klass = Class.new(Celluloid::SupervisionGroup) do
+        def foo
+          add(MyActor, as: :bar, injections: {
+            initialize: Proc.new { @name = :foo }
+          })
+        end
+      end
+      group = group_klass.run!
+      group.foo
+      sleep 0.01
```
digitalextremist: Side discussion -- I keep seeing these. But does anyone think it'd be better to have some sort of `actor.ready?` in any case where race conditions could happen? Ideally adding `group.ready?` here would block until `group` is ready, rather than all this `sleep x` popping up everywhere.

digitalextremist:
This is a very good approach and I am going to merge it immediately so we can double check celluloid/pool#2.. but I am seeing problems with `prepare_options` and getting frustrated enough to do #511 basically immediately.

spec/celluloid/supervision_group_spec.rb
e2: The sleeps need to be fixed - and they are (spec by spec). This one just isn't high priority.

e2:
@digitalextremist - I'd make sure all the tests work first, and the test suite and specs are refactored enough so things like flaky specs and other stuff doesn't get in the way.

The codebase/setup is not ready yet to work on issues like this - besides, I already have extra work merging with the changes related to Pool/Spawner (finishing up just now).

digitalextremist:
Agreed on polishing this prerelease some more before I start on #511. Right now celluloid/celluloid-pool#3 before returning to this prerelease. I am noticing more strange `Pool` behavior than usual.

digitalextremist:
Wait... something you changed fixed it @e2. Any idea what that was?

e2:
@digitalextremist - there are so many things broken to me (unmerged fixes as well), I don't want to even investigate stuff - because it might already be fixed.

digitalextremist:
That's fine. :runner: keep up the awesome and I'll be doing the same.

PR #556, opened 2015-04-08
Rework pool specs

### LINKED ISSUE #554: For review: "truly" synchronous sync calls
TL;DR - seems to work well, but probably has a lot of implications

Here is the patch: https://github.com/e2/celluloid/commit/5e17eae50fc25b7dec3abb0562ee8f36fe781125

I don't particularly like it, but it seemed necessary with Threads, and it seems to help with other JRuby issues.

It bypasses the mailbox when an actor calls it's own methods - this was necessary (in the cases I encountered), because there was no way protect a "receiver" task that makes a sync call to it's sender again (deadlock, because sender is waiting for signal, while receiver mails a sync call to the "waiting" sender task).

On MRI without this line: https://github.com/celluloid/celluloid/blob/master/lib/celluloid/tasks/task_thread.rb#L31 you get a deadlock in the Probe tests. On JRuby though, removing it prevents the following from getting stuck:

```ruby
# Not actual code, but more or less what happens in the actor system:
SupervisionGroup.new do |group|
  group.supervise_as :notifications_fanout, Notifications::Fanout
  # (...)
end
```

Which pretty much becomes:

```ruby
class SupervisionGroup
  def initialize
    supervise_as :notifications_fanout, Notifications::Fanout
  end
end
```

On MRI, the hacky signal "pushes things forward" - but I believe that line is just a deadlock workaround, and not a solution to the sync problem (which doesn't probably exist on MRI anyway - due to how tasks are suspended and resumed everywhere).

I'm still fixing tests, but this seems to work fine. 

The upside may be a small performance increase (no mailbox overhead, etc.), the downside - maybe lower responsiveness in some cases. 

But then again, ideally the user should avoid sync calls unless absolutely necessary (since it potentially means locking the whole call chain).

Would this be ok to merge?

e2:
- made Pool and Spawner have more-or-less the same API
- specs are shared (with slight variations depending on group type)
- catch all Exceptions in Pool/Spawner
- timeouts better help identify the problem
- tests no longer flaky + faster
- include #554 [close #554]


lib/celluloid/groups/spawner.rb
```diff
@@ -19,9 +19,12 @@ def get(&block)
       end
 
       def each
-        threads = []
-        @mutex.synchronize { threads = @group.dup }
-        threads.each { |thread| yield thread }
+        to_a.each { |thread| yield thread }
+      end
+
+      # For temp compatiblity with specs
```
digitalextremist: How temporary is this, because I was definitely trying to avoid having this method.

spec/celluloid/groups/pool_spec.rb
```diff
@@ -1,76 +1,8 @@
...
-
-    it "doesn't leak dead threads" do
-      subject.max_idle = 0 # Instruct the pool to immediately shut down the thread.
-      expect(subject.get { true }).to be_a(Celluloid::Thread)
-
-      sleep 0.01 # hax
-
-      expect(subject.to_a.size).to eq(0)
-    end
+    it_behaves_like "a Celluloid Group"
```
digitalextremist: :+1:

digitalextremist:
With this PR I am seeing dramatic improvements.

lib/celluloid/groups/spawner.rb
e2: Until the specs are "stabily green" at least, and then until I've ported the "fixed" InternalPool - because this API may be unnecessary, but it's good at testing the stuff.

Either way there's probably more refactoring than this - and refactoring is done best when stuff is green - also because while refactoring you might find bugs, so it's best to do it on a "pre-release ready" codebase.

PR #559, opened 2015-04-08
Rework stack dump specs

e2:
- no longer flaky (except because of bugs in Pool and unrelated races occurring on JRuby)
- includes comments about counting threads (and why it makes little sense)
- overall it should be reworked, but other stuff comes first (test framework improvements, Pool fixes, Thread API, shutdown logic, etc.)

### DIFF
--- a/lib/celluloid/calls.rb
+++ b/lib/celluloid/calls.rb
@@ -31,12 +31,7 @@ module Celluloid
       begin
         meth = obj.method(@method)
       rescue NameError
-        inspect_dump = begin
-                         obj.inspect
-                       rescue RuntimeError, NameError
-                         simulated_inspect_dump(obj)
-                       end
-        raise NoMethodError, "undefined method `#{@method}' for #{inspect_dump}"
+        raise NoMethodError, "undefined method `#{@method}' for #{safe_inspect(obj)}"
       end
 
       arity = meth.arity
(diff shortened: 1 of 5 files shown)


digitalextremist:
Watching tests passing.

digitalextremist:
Seems hung up on `Celluloid::StackDump#actors should include all actors` at the very least, but moving forward!

e2:
The last few commits are totally wrong ... I'll fix this.

e2:
The spec is really, really fragile - it really needs an extremely solid foundation (no bugs, no race conditions, etc.).

I've turned of Thread var cleaning for the Spawner. There's no need since the threads are reused - and it can only break code that references stale finished threads.

The spec itself tests too many things at once - though because it's not using mocks, is better to keep it as it is for now (it's especially good for testing changes in the codebase).

Fixing specs like this - before the groundwork is finished - isn't a good idea.

tarcieri:
@e2 yes that spec has traditionally been terrible and also the source of a lot of flaky tests. I'm happy you're even attempting to improve it.

e2:
@tarcieri - thanks. As for flaky tests, pretty much just the Probe specs are left - they trigger the toughest issues with JRuby (too much going on and a different problem every time the tests are run). 

Once they are fixed (on JRuby), it's just a few mop ups here and there and we should be able to get rid of rspec-retry, sleep(), etc.

digitalextremist:
WTF: how are the tests passing? :)

tarcieri:
@e2 the probe specs are likewise the second worst set of specs in Celluloid, so I'm glad you're working on those too :wink:

Issue #463, opened 2014-10-02
Celluloid::Future retaining arguments causing them to not get GC'd


jonhyman:
I'm not sure if this is a bug with Ruby (I'm on ruby 2.1.2p95 but the same behavior exists on 2.0.0.-p247) or Celluloid, but due to a memory leak with Futures we can't use Celluloid at all for our use case because memory can grow unbounded quickly and uses up all the RAM on the machine. I've tested 0.15.2 and 0.16.0 and it appears on both. Here's a very simple repro:

``` ruby
# Setup
require 'celluloid'

def Process.rss ; `ps -o rss= -p #{Process.pid}`.chomp.to_i ; end
def Process.pretty_rss ; Process.rss.to_s.gsub(/(\d)(?=(\d{3})+(\..*)?$)/,'\1,') + " kilobytes" ; end

class Foo
  include Celluloid
  def do_something(arg1)
    return true
  end
end
```

Storing futures in an array, then iterating back over them causes unbounded memory growth:

``` ruby
puts(Process.pretty_rss)
20.times do
  f = Foo.new
  futures = []
  100.times { futures << f.future.do_something("a" * 50_000) }
  futures.length.times {|i| futures[i].value }
  f.terminate()
  futures = []
  puts(Process.pretty_rss)
end

# => 28,384 kilobytes
# => 33,824 kilobytes
# => 39,624 kilobytes
# => 43,396 kilobytes
# => 48,128 kilobytes
# => 53,320 kilobytes
# => 55,900 kilobytes
# => 56,724 kilobytes
# => 59,920 kilobytes
# => 60,844 kilobytes
# => 60,156 kilobytes
# => 64,408 kilobytes
# => 67,520 kilobytes
# => 66,960 kilobytes
# => 70,604 kilobytes
# => 74,948 kilobytes
# => 76,040 kilobytes
# => 75,472 kilobytes
# => 76,280 kilobytes
# => 75,528 kilobytes
# => 79,584 kilobytes
```

If you call `.value` on the future when you make it, growth is contained:

``` ruby
puts(Process.pretty_rss)
20.times do
  f = Foo.new
  futures = []
  100.times { futures << f.future.do_something("a" * 50_000).value }
  f.terminate()
  futures = []
  puts(Process.pretty_rss)
end

# => 80,384 kilobytes
# => 79,728 kilobytes
# => 80,536 kilobytes
# => 79,912 kilobytes
# => 80,720 kilobytes
# => 80,124 kilobytes
[...]

PR #561, opened 2015-04-09
a kind documentation/errata of celluloid-related memory leaks

niamster:


digitalextremist:
@niamster ready to merge this into the prerelease? It looks complete?

e2:
Is there actually a reason to have this test skippable?

niamster:
@digitalextremist yes, it looks complete for me. If I remove 'include Celluloid' from the target class the test is passed with 90% probability.
@e2 it's skipped by default but put into 'allowed failures' for Travis. The reason is that there are still leaks that are linked to ruby itself rather than Celluloid and there's no clean solution to avoid them(see #463)

e2:
@niamster - if I understand correctly, this spec should be enabled on jruby and rbx (because it's expected to pass) and it should be always disabled on MRI (until there's a workaround).

Either a test can pass 100% of the time - and should be enabled (on whatever platforms support it), or disabled completely (until it works 100% of the time on MRI as well).

I think of this as a test protecting against *introducing* certain memory leaks - and so it should enabled everywhere it can.

Also, think of JRuby (and RBX) as the *main* interpreters for Celluloid. MRI is the one that should have special patches to match the behavior of the other two.

digitalextremist:
I'm a bit off radar today but just wanted to say: I believe it's the opposite @e2. MRI isn't leaking and jRuby / rbx are. So those will fail unless @niamster figured out how to work around this. I saw `Fiber.yield` come in, so maybe he did!

niamster:
@digitalextremist I didn't try these tests with rbx and jruby.

niamster:
Going to push new one with a changeset which fixes one of the leaks in MRI.

Issue #420, opened 2014-05-20
Celluloid::Logger breaks existing code


ioquatix:
Code expecting to get `::Logger` ends up getting `Celluloid::Logger` in specific contexts where celluloid has been included. This behaviour is unexpected.

https://github.com/celluloid/celluloid-io/issues/107

digitalextremist:
@ioquatix do you have this covered, and if so are you able to get to it in the next few weeks?

ioquatix:
This is still an issue.

digitalextremist:
Yep. Working on it //

e2:
:+1: - I wanted to suggest "internals" exactly myself. 

I'd also considered namespacing them though to match, e.g. `Celluloid::Internals::Foo` - a bit more complex in code, but it's also a clear distinction for contributors as to what constitutes "internals" and not.
Also, `Internals::Foo` isn't that bad/long really.

digitalextremist:
@e2 that's how it's happening, right after `Celluloid::FSM` is out ( moments from now ) ... since that was so obvious and easy to ship.

ioquatix:
Awesome :)

digitalextremist:
Fixed at cfd06898ccbbe09cf295e63e92633c78ca54e255

PR #569, opened 2015-04-11
`Internals` refactoring branch on top of 0.17.0 release

digitalextremist:


e2:
@digitalextremist - let me know once you're done with renaming files/classes. There's stuff like the InternalPool reappearing (in internals), uuid appearing twice, registry, etc.

I don't mind the rebasing that much - just that there will be extra confusion if we both go about fixing the same issues (and in different ways).

E.g., I'm currently implementing Celluloid::Reaper for complex/extensible shutdown logic (it will let you close a few Celluloid issues and is designed to be used by Supervisor trees to shutdown their actors), the (Internal)Pool is now working again (but before these recent renames), and I'm removing Supervisors from non-supervisor tests (to help you with the Supervisor refactoring/extracting). 

And there are still some issues with tests failing every now and then - and I need a "stable base" to properly track down whether the issues are related to tests, interpreters or bad bugs/races in Celluloid.

Just giving you a heads up so we don't get in each others way.

If I find any "major" stuff, I'll put those on a separate branch - so you can merge it into 0.17.x without the baggage of experimental/development stuff you won't need.

Keep up the great work!

digitalextremist:
@e2 I'll be done renaming today most likely, then I'll signal. The issues to resolve for #420 are a little weird. I'm aware of the mess/duplication. Thanks for fixing that `Yield` error. That in and of itself brought a lot of sanity back to the prerelease home-stretch.

Would `Reaper` be a `Group` subclass? The name connotes garbage collection to me, maybe you're looking more toward `Shutdown` or that sort of name?

What do you think about extracting `Probe` into a gem?

e2:
@digitalextremist - I want it to be a separate class simply because it can be more easily tested (and extended). Basically, it's a ActorSystem#shutdown reimplementation, though it has a lot in common with the new (Internal)Pool shutdown as well - so in a way, it is related to Group.

But I'd keep it separate, because it shouldn't share state with any other class - it needs to be thoroughly tested. The last thing I'd want is to have to setup the full Celluloid stack for coverage on every 'if' condition, and then loads of code to avoid raciness and flakyness.

Also, I'd like to keep it as isolated from everything else as possible - because I doubt the initial design will survive the needs people will have. (If we tie it to other classes it will be hard to fix without breaking compatibility, etc.). It's also noticeably slower, because it e.g. actually allows actors to process TerminationRequests (instead of politely telling them to terminate and then immediately murdering them anyway).

Also, existing code should be "disabled" to some extent. Things like "kill" and "join" should either go through the Reaper, or be unavailable - due to deadlocks for example. (The only reason you don't get deadlocks on MRI during shutdown is because the whole method is wrapped in a Timeout block).

I'd put it in internals, e.g. Celluloid::Internals::Reaper - unless you have a better idea/suggestion for a name.

> Thanks for fixing that Yield error. 

I'm happy I got to add a test for such a critical path. I'm more worried about the Probe tests, because they could be a bad bug lurking in Celluloid - something worth squashing before the release.

> What do you think about extracting Probe into a gem?

I think tracing should be implemented. Probe isn't that good, since the events fired don't provide much info. Also Probe is kind of married to Celluloid ATM, because of the $CELLULOID_MONITORING at critical places. 

Tracing would solve both the Probe and debugging needs at the same time (and it would allow refactoring out logging calls from deep within Celluloid). It would also allow easier debugging of production apps, where you can't really just change the log level and browse through megabytes of logs (and mostly containing "dangerously suspending" warnings I guess...).

But that's far off. Overall, Probe is probably not worth extracting out. It's not that heavy, but it's deeply hooked in, with now capability in Celluloid to turn it into a seamless "plugin".

Probe has 3 problems, really:
1. the events have no info (e.g. `:actor_named` without saying what the new registry name is)
2. problematic, complex and flaky tests (hard to debug partially because of 1)
3. shutdown sequence (I don't know if it's even possible to set it up so there's no crash - that's where I think the Reaper combined with a supervision tree would solve the problem)

Issue #564, opened 2015-04-11
Extract FSM into gem.


digitalextremist:
With a simple search of the codebase, `Celluloid` itself isn't even using FSM. Only specs and examples use it. This seems natural to extract, for clarity.

e2:
:+1:

digitalextremist:
This is passing all tests, live at [celluloid/celluloid-fsm](/celluloid/celluloid-fsm), and cut onto rubygems.

digitalextremist:
It could use some cleaning though, probably.

ioquatix:
Nice work everyone.

PR #570, opened 2015-04-11
branch coming in over the top of pre-release before massive review/clean up

digitalextremist:
Will be handling #420 ( already in progress ) and handles #564, and parts of #511.

Definitely a work in progress, starting to regain traction in overall release.


digitalextremist:
Although this is *right in the middle* of #420 this has gotten too fractured so I'm bringing this back into the prerelease branch and going to sort that out versus two concurrent prerelease branches for the gems to keep track of.

PR #580, opened 2015-04-15
misc spec fixes

e2:
@digitalextremist - based on what I checked, this won't destabilize 0.17.0 (as opposed to experimental branch).

Lots of fixes and improvements for specs and development.

(Extremely non-invasive changes to Celluloid itself).

I've refreshed RuboCop config and explicitly made just double quotes mandatory.

(The setup allows gradually re-enabling offenses without disrupting work, introducing merge problems, etc. - while preventing new PRs from accidentally introducing new offenses).


tarcieri:
Nice!

digitalextremist:
Going through the entire PR now. Literally fell asleep under my desk.

lib/celluloid.rb
```diff
@@ -478,17 +478,30 @@ def future(meth = nil, *args, &block)
...
 
-Celluloid.group_class = begin
-  Celluloid::Group.const_get(ENV['CELLULOID_GROUP_CLASS'] || fail(TypeError))
-rescue
-  Celluloid::Group::Spawner
-end
+Celluloid.task_class =
+  begin
+    str = ENV['CELLULOID_TASK_CLASS'] || 'Fibered'
+    Kernel.const_get(str)
```
digitalextremist: I wanted to *force* `Celluloid::Group` or `Celluloid::Task` as the base for either class type. Might seem petty, but this convention is a stern one. If it's really a hangup for people, they can override it in Ruby code directly.

e2: The thing is: other people actually shouldn't follow the convention. One example is stubbing - e.g. MySpecs::SerialPool should be acceptable. Forcing the constant to belong to Group is only necessary if you're iterating through the available groups ... which probably doesn't make sense in the context of Celluloid.

I'd say it's actually better to *not* allow people to use that namespace - then, if something ever crashes, you can have Celluloid blame the user if the class isn't in the Celluloid namespace (meaning - it's not a class that's part of Celluloid's "approved" classes).

Here, though, what's important for me is to get an exception if I mistype 'Threaded', because suddenly the tests will be using Fibered ... and they can pass. So I want it to fail if my class doesn't exist (e.g. when you switched the names, my tests started succeeding when they shouldn't).

digitalextremist:
@e2 seems like you're still adding to this so let me know when it can be merged

e2:
@digitalextremist - just fixing new failures on Travis (it's not green yet). 

The crash detection turns crashes into actual tests failures. In this case (last commit), this overlaps with the Reaper (already fixed there), but the Reaper may not be ready for 0.17.0 (needs lots of work).

Meanwhile, this is a cleanup bug - one task can prevent other tasks from being terminated.

e2:
@digitalextremist - all green (all platforms), so I'm done. If you encounter a single test failure at any time with this, just report it as a new issue and cc me.

PR #584, opened 2015-04-16
memory leak battle

niamster:
changes:
 * rebase onto the tip of 0.17.0-prerelease
 * fixed memleaks:
  - method args and return value on MRI/2.2
  - method return value in jruby
  - get rid of object space module as it's not supported in rbx or jruby


niamster:
Argh, a typo after rebase.

e2:
@niamster - great stuff. 

I find the mailbox code quite confusing with the timers, checking, locks, receive()+check() API ... and that break returning a message (which was a break from the timer loop).

niamster:
@e2 not so tricky IMHO. What I hate here is that we have to violate nice ruby syntax to avoid memory leak =|

tarcieri:
@niamster no worries about syntactic beauty. Make it work then make it beautiful :wink:

digitalextremist:
@niamster the followup #585 to this was *functionally* beautiful. It has really helped with the fractures I've been seeing. Celluloid is starting to behave like itself again.

PR #585, opened 2015-04-16
memory leak battle

niamster:
changes:
   - rebase onto the tip of 0.17.0-prerelease
   - fixed memleaks:
     - method args and return value on MRI/2.2
     - method return value in jruby
     - get rid of object space module as it's not supported in rbx or jruby


digitalextremist:
:+1: merging

digitalextremist:
@niamster this PR has dramatically improved the stability of celluloid from what I've so far seen. Thank you for your work on this!

niamster:
@digitalextremist really? That's interesting. Anyway I will continue to work especially on the leak of the actors.

marshall-lee:
@niamster could you explain these MRI- and JRuby-specific issues? Very interesting!

digitalextremist:
In my understanding, #463 goes deeply into that @marshall-lee.

PR #587, opened 2015-04-17
remove dependency on supervisors

e2:
@digitalextremist - this removes Celluloid's dependency on supervisors (Fanout and IncidentReporter will only be supervised if supervisors are included/available).

Still, supervisors will have to be adapted I guess.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,12 +4,6 @@ gem 'coveralls', require: false
 
 gem 'timers', github: 'celluloid/timers'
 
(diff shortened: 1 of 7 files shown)


e2:
@digitalextremist - the celluloid-supervision specs pass after this patch.

digitalextremist:
We decided to extract `celluloid-essentials` and keep supervision.

e2:
@digitalextremist - why keep supervision (asking because the code here also removes unnecessary supervision usage in the many actor_system-dependent specs).

PR #594, opened 2015-04-19
spec: use 'skip' rather than 'pending' as rspec still tries to execute the case and marks it as failed

niamster:
The test case is failed when actor spec example is used in celluloid-zmq or -io


tarcieri:
Can't you just use `xit` for this?

niamster:
It works either way. You prefer `xit`?

tarcieri:
Seems simpler *shrug*

PR #595, opened 2015-04-19
spec: use 'xit' rather than 'pending' as rspec still tries to execute the case and marks it as failed

niamster:


tarcieri:
Scumbag Coveralls. I just disabled the status API so that shouldn't happen again.

e2:
@niamster - this is a feature of RSpec. 

The "pending" marks tests as failed (and runs them so to give an overview why they fail), but they don't influence the build exit code (it succeeds) - *but*, if the "pending" tests actually succeed, this fails the build (because it's an error if you have a pending test that somehow passes).

The best workaround (to avoid running the broken tests if they create side effects) is to just `fail "TODO"` under the pending.

e2:
I added the Celluloid::IO fix (to make the mocking tests fail/pass both in Celluloid::IO and Celluloid identically) here: https://github.com/celluloid/celluloid-io/pull/134

PR #599, opened 2015-04-21
mailbox bypass and no meth checking

e2:
@digitalextremist - let me know what issues you're having.

(NOTE: I also reverted the sync calls short-circuit for this to work).

### DIFF
--- a/lib/celluloid/calls.rb
+++ b/lib/celluloid/calls.rb
@@ -22,7 +22,6 @@ module Celluloid
     end
 
     def dispatch(obj)
-      check(obj)
       _b = @block && @block.to_proc
       obj.public_send(@method, *@arguments, &_b)
 =begin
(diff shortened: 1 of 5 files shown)


digitalextremist:
I am open to approaching this as an experimental gem if you are interested. It could be a testing ground for concepts. What it'd end up being is a third mode, so there'd be `:backported`, `:current`, and `:experimental`. Only objects and methods that override `:current` would need to be included.

e2:
@digitalextremist - currently (without this patch), the more methods an actor has, the slower it is (which discourages refactoring complex methods in actors). 

OTOH, by calling every method through the mailbox, latency may be improved (except on MRI) if threads are scarce - but the best solution to latency is to use multiple actors (it's faster). But, having methods go through the mailbox can create unexpected hard-to-debug side-effects (if a user's methods are called out of order).

Summary: in production, this patch speeds up execution and prevents unexpected side-effects - so even if it was marked as "experimental", I'd use it by default myself.

digitalextremist:
Then, yes... gem? Or, branch?

e2:
@digitalextremist - gem of course.

digitalextremist:
@e2 you gave no indication whether gem or branch in your reply. Ok, then.

At abstractive labs: [celluloid-experimental](/abstractive/celluloid-experimental)
Skeleton gem: https://rubygems.org/gems/celluloid-experimental

You have write access to the repository at abstractive labs. I am closing this here then, and suggest you move all experimental issues/PR's to that repository. I am eager to test some of these concepts you've been talking about and see them in action.

PR #602, opened 2015-04-22
Update supported Rubies

tarcieri:
1.9.3 is past EOL

2.2.2 is the latest and we should be checking it on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 script: rake ci
 rvm:
-  - 1.9.3
   - 2.0.0
   - 2.1.4
+  - 2.2.2
   - ruby-head
   - jruby
   - jruby-head
@@ -10,7 +10,6 @@ rvm:
 
 matrix:
   allow_failures:
-    - rvm: 1.9.3
     - rvm: ruby-head
     - rvm: jruby
     - rvm: jruby-head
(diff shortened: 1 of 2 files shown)


digitalextremist:
This seems good to go? This was basically done on the next release, but not here on master.

tarcieri:
When you say "good to go", you mean unnecessary, or mergeable?

digitalextremist:
I mean mergeable, or do you have more you were going to add?

tarcieri:
Nope! Merged

digitalextremist:
>> right on <<

kostya:
you can set just 2.1, 2.2 and travis get last minor version

digitalextremist:
@kostya I don't see anything mentioning this behavior on Travis CI, in their examples they specify a minor version number. But also, we tend to need to test point release by point release, so it seems like it would be better to manually specify which minor version number. We've even marked individual point releases as allowable failures since certain point releases sometimes break ( without Celluloid itself being broken ). Make sense?

kostya:
this is rvm behavior, from 2.1, minor version is just patch level.

tarcieri:
@kostya if I use `2.2` in `.travis.yml` it selects 2.2.0, not 2.2.2

kostya:
hm it works with 2.1, and should work with 2.2 too, but travis used not last dvm :(

PR #605, opened 2015-04-26
Logical && and || operators for simple Boolean logic

olleolleolle:
Where there is no control-flow trickery, only simple if/unless statements, we can use `&&` and `||`.

[Avdi Grimm made a fine screencast on the and/or feature](http://devblog.avdi.org/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/) and how best to use it.

One more thing: what are the possible values of `CELLULOID_FUTURE`? Did it ever get set? It was introduced in this file in 2013. Can not find reference to it in this repo.

### DIFF
--- a/lib/celluloid.rb
+++ b/lib/celluloid.rb
@@ -10,12 +10,12 @@ require 'celluloid/notices'
 
 $CELLULOID_BACKPORTED = false
 
-if ENV['CELLULOID_BACKPORTED'] == 'silently' or ( defined? $CELLULOID_BACKPORTED and $CELLULOID_BACKPORTED == :silently )
+if ENV['CELLULOID_BACKPORTED'] == 'silently' || ( defined?($CELLULOID_BACKPORTED) && $CELLULOID_BACKPORTED == :silently )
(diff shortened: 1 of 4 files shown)


tarcieri:
I agree with this change

digitalextremist:
Thanks @olleolleolle! When I came over from C++ style languages I assumed the English keywords were replacements for `&&` and `||` but after the excellent screencast, I see they are not at all intended for conditional statements primarily, but for flow control. It's nice to learn something so commonly used.

Looks like you've been reviewing `0.17.0`. I appreciate that.

`CELLULOID_FUTURE` itself can probably be depreciated. I believe merely setting it ( not setting a certain value ) triggered the equivalent of `$CELLULOID_BACKPORTED == false`

digitalextremist:
@tarcieri can you think of any instances where `CELLULOID_FUTURE` was promoted to the community, or any obvious uses out in the wild? I'm thinking that I will advocate for using `require "celluloid/current"` for the brave/serious sorts who refactor, rather than ever suggesting someone sets a global variable.

PR #608, opened 2015-04-27
Example for repeating timer with #every

olleolleolle:
For beginners: an example of `#every`.

 - prints a single line of consecutive numbers (simulating going to sleep counting sheep)
 - added a puts at the end to avoid messing up last line of output
 - the added example is delineated from the rest of the example with a clear class name marker in the output

### DIFF
--- a/examples/timers.rb
+++ b/examples/timers.rb
@@ -45,3 +45,20 @@ actor.timer.cancel
 # Wait until after the timer should've fired
 sleep 3.1
 puts "Timer shouldn't have fired because we cancelled it: #{actor.fired}"
+
+class RepeatingTimerExample
+  include Celluloid
+
+  def count_sheep
+    print "<#{self.class.name}> Counting sheep to go to sleep: "
+    sheep = 0
+    every(0.1) do
+      sheep += 1
+      print sheep, ' '
+    end
+  end
+end
+
+RepeatingTimerExample.new.count_sheep
+sleep 1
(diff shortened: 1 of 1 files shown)


digitalextremist:
Good idea! And would you please add `#stop_counting` after the `sleep 1` to illustrate canceling timers, rather than having the death of the main thread end the timer?

You'd just need to store the timer, then cancel it. You could even just `def_delegator` the `cancel` method to an `@timer` instance variable. You know what I mean?

olleolleolle:
(Readability: Focusing on the timer functionality, I skipped the delegation. Is that good/bad?)

digitalextremist:
Oh that's totally fine @olleolleolle, thanks for adding that.

digitalextremist:
I'll merge this in probably in 24-48 hours, before cutting th RC. Thanks again for the example.

digitalextremist:
Thanks @olleolleolle. This is merged, and I corrected two things: `celluloid/current` and using `@sheep`

olleolleolle:
Just pointing out that the attr_accessor allows the previous usage. Now, the attr_accessor :sheep is superfluous.

digitalextremist:
Yeah, I had seen that but strangely it was still erroring out.

digitalextremist:
Ah, because it's being incremented inside a timer block. Withdrawing `attr_accessor`

PR #604, opened 2015-04-30
Execute block on receiver together with exclusive mode

Burgestrand:
Executing the block on the receiver is essentially passing the block as-is without wrapping it in Celluloidy magic to pass execution back and forth. I was expecting to be able to pass a block to a method and execute it on the receiver during exclusive mode, but I was not.

```ruby
require "bundler/setup"
require "celluloid"

Celluloid.logger = nil

class A
  include Celluloid

  def exclusive_on_receiver
    b = B.new
    exclusive { b.on_receiver { $stderr.puts "OK" } }
  end
end

class B
  include Celluloid

  execute_block_on_receiver def on_receiver(&block)
    yield
  end
end

begin
  A.new.exclusive_on_receiver
rescue => error
  $stderr.puts "ERROR: #exclusive_on_receiver (#{error})"
end
```

I understand executing blocks during exclusive mode could cause deadlock issues, but that is true for anything performed under exclusive mode. Is there some other reason that this does not work that I am not seeing?

### DIFF
--- a/lib/celluloid/call/sync.rb
+++ b/lib/celluloid/call/sync.rb
@@ -28,6 +28,18 @@ module Celluloid
         Internals::CallChain.current_id = nil
       end
 
+      def check(obj)
+        super
+
+        if task && block && block.execute_on_sender? && task.exclusive?
+          raise "Cannot execute blocks on sender in exclusive mode"
+        end
+      rescue AbortError
+        raise
+      rescue => ex
+        raise AbortError.new(ex)
+      end
+
       def cleanup
         exception = DeadActorError.new("attempted to call a dead actor")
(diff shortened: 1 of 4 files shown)


Burgestrand:
`export CELLULOID_BACKPORTED=false` did the trick nicely. The deadlock I found had nothing to do with Celluloid, I managed to try to run an expectation on a recursive list, turns out that does not work :bomb:.

All tests pass on my end now, and I believe the code is ready, so I've turned this issue into a pull request.

digitalextremist:
@Burgestrand can you please rebase this and resubmit it against the `0.17.5-release` branch?

Burgestrand:
@digitalextremist sure, I do believe so.

PR #616, opened 2015-05-08
Update pubsub.rb

scorix:
fix `NoMethodError: undefined method 'publish' for Celluloid::Notifications:Module`


digitalextremist:
How are you even getting this error? `Celluloid::Notification.publish` does exist, and properly calles `.notifier` on `master` ... and in the current upcoming release, all this moved to the `celluloid-essentials` gem.

PR #647, opened 2015-07-07
Execute block on receiver together with exclusive mode

Burgestrand:
This has been rebased on 0.17.5-release. See #604 for previous discussion.

### DIFF
--- a/lib/celluloid/calls.rb
+++ b/lib/celluloid/calls.rb
@@ -8,10 +8,6 @@ module Celluloid
       @method = method
       @arguments = arguments
       if block
-        if Celluloid.exclusive?
-          # FIXME: nicer exception
-          fail "Cannot execute blocks on sender in exclusive mode"
-        end
         @block = Proxy::Block.new(Celluloid.mailbox, self, block)
       else
         @block = nil
--- a/lib/celluloid/calls.rb
+++ b/lib/celluloid/calls.rb
@@ -19,7 +19,12 @@ module Celluloid
     end
 
     def dispatch(obj)
-      check(obj)
+      begin
+        check(obj)
+      rescue => error
+        raise AbortError.new(error)
+      end
+
       _b = @block && @block.to_proc
       obj.public_send(@method, *@arguments, &_b)
       #     rescue Celluloid::TimeoutError => ex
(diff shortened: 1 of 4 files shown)


digitalextremist:
@Burgestrand thank you very much for doing this!

ioquatix:
Can we merge this?

lib/celluloid/calls.rb
```diff
@@ -23,9 +15,13 @@ def execute_block_on_receiver
     end
 
     def dispatch(obj)
-      check(obj)
-      _b = @block && @block.to_proc
-      obj.public_send(@method, *@arguments, &_b)
+      begin
+        check(obj)
+      rescue => error
+        raise AbortError, error
```
tarcieri: This seems broken to me. It means that the `ArgumentErrors` below will only crash the receiver (but not the caller?)

ioquatix: Could you add a test to cover this behaviour so we can rebase and see it break?

Burgestrand: Yes, will do. This does indeed seem broken.

Burgestrand: From what I can see this is already being tested here: https://github.com/elabs/celluloid/blob/0.17.5-release_allow-receiver-blocks-in-exclusive-mode/spec/celluloid/calls_spec.rb — ArgumentError and NoMethodErrors properly crashing the sender and not the receiver. If I recall correctly (it's been a few months, memory rusty), the AbortError is specifically to crash the sender and not the receiver. Catching the errors when doing check, as opposed to inside the check-method, should make no difference.

Furthermore I tried rebasing this on 0.18.0-prerelease, but I had some conflicts so I stopped that endeavour.

Finally, I'm sorry, but I believe I'll personally be dropping this issue. Thanks for the work you've put into all of this, but at least for the coming few weeks/months I'm done.

ioquatix:
Shouldn't we merge this?

tarcieri:
It needs a rebase to be mergeable. I can push it back up and reopen the PR.

ioquatix:
Ah yep, perhaps I will have time next week I will take a look if nothing has happened by then.

PR #649, opened 2015-07-10
Docs, minor clarification.

neektza:
Clarification added that explains that the `Actor#handle` method registers handlers for a given pattern.


digitalextremist:
Merged, thanks @neektza

PR #653, opened 2015-08-04
Start of 0.17.5 release

digitalextremist:
Going to do this release in `master`, with each piece passing as it's pushed in... rather than a huge waterfall PR at the end. This needs to go in to stop all the gems from breaking, since there is a breaking change here... and deprecation patches that are needed as well.


digitalextremist:
All mandatory tests passing. Moving into master.

PR #654, opened 2015-08-05
Refactor (Standard)Errors vs. (Signal)Exceptions

### LINKED ISSUE #424: Celluloid::Task::TerminatedError really StandardError?
Stopping an actor seems to sometimes result in Celluloid::Task::TerminatedError - but this often ends up going through the registered exception handlers since it is derived from StandardError. Wouldn't it make more sense for this to be derived directly from Exception and perhaps called "TerminatedException" - it isn't really an error condition right?

digitalextremist:
Closes #424

Introduces varies of `*::TimedOut` which along with `TaskTerminated` are `Celluloid::Interruption` subclasses, deriving from `SignalException` ( seen as an in-process `Interrupt` rather than an "Error" )

/cc: @ioquatix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ rvm:
 matrix:
   fast_finish: true
   allow_failures:
-    - rvm: rbx-2
     - rvm: ruby-head
     - rvm: jruby-head
     - env: CELLULOID_BACKPORTED=true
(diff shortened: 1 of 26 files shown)


ioquatix:
This looks like a good change, I'd just be careful about creating a larger surface area if it isn't necessary, so I'd try to keep any new exceptions classes to a minimum :)

digitalextremist:
@ioquatix I think what you might be seeing in terms of "new errors" and nesting, is that (here too) I felt the same and I've tried to collect together and reassess whether we need layers of errors elsewhere in the system, and consolidated them in `exceptions.rb` ... you caught some I didn't thin out, but otherwise I think we've *lost* many versus gained more.

ioquatix:
Awesome, thanks for fixing this.

digitalextremist:
My pleasure. Feels cleaner.

PR #662, opened 2015-08-11
remove requiring current and backported version from readme

fbehrens:
in current celluloid 0.16.0 both files are not exsisting anymore.


digitalextremist:
The current version is `0.17.1.1`

PR #665, opened 2015-08-17
Advise to load submodules for using pre-release

marshall-lee:
Without `submodules: true` one would get this error:
```
Fetching git://github.com/celluloid/celluloid.git
There was a LoadError while loading celluloid.gemspec: 
cannot load such file -- /home/vkochnev/.rvm/gems/ruby-2.2.1/bundler/gems/celluloid-995589fc28a2/culture/sync
from
  /home/vkochnev/.rvm/gems/ruby-2.2.1/bundler/gems/celluloid-995589fc28a2/celluloid.gemspec:3:in `<main>'

Does it try to require a relative path? That's been removed in Ruby 1.9.
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ If you don't get anywhere with this checklist, please feel free to [file a bug r
 If you encounter a bug, it's entirely possible that it has already been fixed but not yet included in a released version. You can establish this by trying to run your application with a pre-release version of Celluloid direct from source control. You can do this by modifying your application's Gemfile as follows:
 
 ```ruby
-gem 'celluloid', github: 'celluloid'
+gem 'celluloid', github: 'celluloid', submodules: true
 ```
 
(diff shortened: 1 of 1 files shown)


digitalextremist:
Thanks for catching that in the docs.

The requirement is not a relative path, but a path that does not exist without the submodule's files.

This will be forever solved once we bring together our repositories as one omnibus repository, because the submodule right now exists to synchronize the dispersed repositories only.

marshall-lee:
@digitalextremist 

Another question: how to open PRs to celluloid properly? My first commit here was without `[ci skip]` and build failed because of `timers` dependency.

digitalextremist:
Generally that does not happen. You've done everything properly. I am looking into why `timers` is complaining.

digitalextremist:
I recut nearly every gem with the new `celluloid-culture` values; specifically `timers` pinned at `>= 4.1.1` versus `~> 4.0.0` and it is now resolved. Everything you did seemed right.

PR #666, opened 2015-08-17
Don't catch IOError

marshall-lee:
Evented mailbox should not know anything about IOError.

Some notes:

1) Method `<<` is changed to catch any error that `wakeup` can raise. It can be not only `IOError`.
2) Method `check` is more interesting. It just converts `IOError` to `MailboxShutdown` and I think it's a cryptic behavior. It is also very related to https://github.com/celluloid/celluloid-io/issues/151. `Celluloid::IO`'s reactor calls `task.resume` inside `run_once`. This can lead to the situation when `IOError` is caught by mistake (raised by user's code for example or by socket operation). So `MailboxShutdown` should be raised by *reactor* itself but not by generic `Mailbox::Evented` (if there's really any need in this exception — now it's indistinguishable from `MailboxDead` in its behavior) .

### DIFF
--- a/lib/celluloid/mailbox/evented.rb
+++ b/lib/celluloid/mailbox/evented.rb
@@ -31,7 +31,7 @@ module Celluloid
         begin
           current_actor = Thread.current[:celluloid_actor]
           @reactor.wakeup unless current_actor && current_actor.mailbox == self
-        rescue IOError
+        rescue
           Internals::Logger.crash "reactor crashed", $ERROR_INFO
           dead_letter(message)
         end
@@ -51,8 +51,6 @@ module Celluloid
 
         # No message was received:
         return nil
-      rescue IOError
-        raise MailboxShutdown, "mailbox shutdown called during receive"
       end
 
       # Obtain the next message from the mailbox that matches the given block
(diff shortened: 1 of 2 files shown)


marshall-lee:
How to fix a build? It fails on initialization phase:

```Could not find gem 'timers (~> 4.0.0) ruby' in

git://github.com/celluloid/timers.git (at master).

Source contains 'timers' at: 4.1.0```

**Comment by marshall-lee** [comment]
@digitalextremist build is okay now. what do you think?

**Comment by marshall-lee** [comment]
Rebased.

**Comment by marshall-lee** [comment]
Simple explanation of this bug:

```ruby
class Worker
  include Celluloid::IO

  def perform
    # if I raise IOError here it will be dispatched in Celluloid.
    # If Worker is supervised it normally restarts after this.
    @socket.read
    # ... task is resumed by Celluloid::IO::Reactor
    # if I raise IOError here it will be dispatched in Celluloid::Mailbox::Evented.
    # If Worker is supervised it doesn't restart.
  end
end
```

ioquatix:
Can we go ahead and merge this? It looks good.

pboling:
:+1: based on the description of the bug, I need this for my supervised workers I use with my [celluloid-io-pg-listener project](https://github.com/pboling/celluloid-io-pg-listener).

PR #678, opened 2015-09-20
Clean finished threads in Group::Spawner.

### LINKED ISSUE #670: Memory Leak using Celluloid::Future
I am using `Celluloid::Future` to kick off tasks and delay their execution until the result is requested. In version 0.17.0, I am finding that this is resulting in an apparent memory leak. Each call I run appears to create a new thread in the `Celluloid.actor_system.group.group` array. The `Celluloid::Group::Spawner` appears to never clean up the completed Threads that it creates.

As a work around, I am planing to instead configure my applications to use the `Celluloid::Group::Pool` group class which appears to support proper thread pooling. However in digging through this, I noticed a very aggressive warning at the top of this file (https://github.com/celluloid/celluloid/blob/master/lib/celluloid/group/pool.rb#L6), which raised a healthy level of concern for me. Don't hesitate to let me know if there is an alternative way to work around this instead. Thanks.

marshall-lee:
This is my attempt to fix #670

### DIFF
--- a/lib/celluloid/group/spawner.rb
+++ b/lib/celluloid/group/spawner.rb
@@ -17,18 +17,18 @@ module Celluloid
 
       def shutdown
         @running = false
+        queue = []
         @mutex.synchronize do
-          queue = Queue.new
           loop do
             break if @group.empty?
             th = @group.shift
             th.kill
             queue << th
           end
-          loop do
-            break if queue.empty?
-            queue.pop.join
-          end
+        end
+        loop do
+          break if queue.empty?
+          queue.pop.join
         end
       end
 
@@ -59,6 +59,7 @@ module Celluloid
               Thread.current[:celluloid_meta][:state] = :finished
             end
             Thread.current[:celluloid_meta][:finished] = Time.now
+            @mutex.synchronize { @group.delete Thread.current }
           end
         end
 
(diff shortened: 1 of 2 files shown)


digitalextremist:
I will review this and test it. Thanks for the PR here.

digitalextremist:
So far I am very happy you patched this, and like what I see. My only concerns are to make sure the test is sufficient to cover all scenarios, which on first glance it seems to be... and wondering if the cleanup at shutdown ought to happen inside the `synchronize` block or not. I will merge this and make any needed tweaks but otherwise the behavior seems correct.

marshall-lee:
Some comments on why I had to change `shutdown` code exactly this way:

- removed `Queue` usage in favor of simple array because there's no inter-thread communication in which `Queue` is helpful.
- extracting `join` call out of `synchronize` there were two reasons: 1) there's simply no need in this since `@mutex` is aimed to guard only `@group`  destruction. 2) since i used `synchronize` inside thread too there's a possible deadlock calling `join` on this thread inside of another `synchronize` — and even some tests with `sleep` were failing until I resolved this deadlock.

digitalextremist:
I thought you might be avoiding mutex deadlocking, understood.

digitalextremist:
There *is* a race condition in the test, on `rbx`

marshall-lee:
@digitalextremist I can't reproduce it on my computer but I understand how it's possible. Attempted to fix it in #679 
Sorry for inattention, I should had write this spec just like the rest in `spec/shared/group_examples.rb`.

digitalextremist:
Your #679 PR is perfect. Thank you

PR #679, opened 2015-09-21
Fix Group::Spawner spec

marshall-lee:


digitalextremist:
Exactly the solution I thought best also. Thank you. Once the tests pass I will merge this also.

PR #682, opened 2015-10-04
Remove excess call/block require

gazay:
`celluloid/call/block` required twice.

Btw why `Call` class file named `calls`? Maybe better to rename it to `call.rb`?


ioquatix:
I guess if anything, extracting `Celluloid::Call` into it's own file and leaving calls.rb to include everything might be a useful cosmetic improvement.

PR #686, opened 2015-10-22
Print out method name to help debugging method call which caused dead…

ioquatix:
… actor error.


digitalextremist:
Looks good to me :+1:

( Only wonder if maybe the `#` character ought to be inserted before the method name, to make it clearer it's a method name? )

PR #688, opened 2015-11-04
Fix failure to remove dead actors from sets, e.g. celluloid-supervision.

ioquatix:
Discussed here:

https://github.com/celluloid/celluloid-supervision/issues/26

This is a pretty important bug fix, it might even warrant being back ported since it's currently breaking our production code.

### DIFF
--- a/lib/celluloid/proxy/abstract.rb
+++ b/lib/celluloid/proxy/abstract.rb
@@ -6,7 +6,7 @@ class Celluloid::Proxy::Abstract < BasicObject
   end
 
   # Needed for storing proxies in data structures
-  needed = [:object_id, :__id__, :hash, :private_methods] - instance_methods
+  needed = [:object_id, :__id__, :hash, :eql?, :private_methods] - instance_methods
   if needed.any?
     include ::Kernel.dup.module_eval {
       undef_method(*(instance_methods - needed))
--- a/spec/shared/actor_examples.rb
+++ b/spec/shared/actor_examples.rb
@@ -29,6 +29,17 @@ RSpec.shared_examples "a Celluloid Actor" do
   it "can be stored in hashes" do
     expect(actor.hash).not_to eq(Kernel.hash)
     expect(actor.object_id).not_to eq(Kernel.object_id)
+    expect(actor.eql? actor).to be_truthy
+  end
+
+  it "can be stored in hashes even when dead" do
+    actor.terminate
+    
+    expect(actor.dead?).to be_truthy
+    
+    expect(actor.hash).not_to eq(Kernel.hash)
+    expect(actor.object_id).not_to eq(Kernel.object_id)
+    expect(actor.eql? actor).to be_truthy
   end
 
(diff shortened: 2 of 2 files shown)


ioquatix:
cc @tarcieri @digitalextremist

pboling:
CC @rgtru @brettfishman I think we'll need this too.

ioquatix:
The more people that review this probably the better. Can everyone have a good critical look at the code and let me know what you think? It's only a small amount but it's at a very important part of Celluloid..

lib/celluloid/proxy/actor.rb
```diff
@@ -19,6 +19,14 @@ def inspect
     "#<Celluloid::Proxy::Actor(#{@mailbox.address}) dead>"
   end
 
+  def eql? other
+    @mailbox.eql? other.mailbox
```
tarcieri: I'd suggest checking `class` too, otherwise all subclasses of proxy will compare equal...

ioquatix:
@tarcieri do you mean `__class__`?

tarcieri:
Yes

ioquatix:
@tarcieri do you think it makes more sense to define a working `kind_of?` / `is_a?` on AbstractProxy?

tarcieri:
@ioquatix absolutely not for `eql?`. You should ensure the classes are identical

ioquatix:
@tarcieri my question was if it makes sense to define `kind_of?` / `is_a?` but I wasn't really considering that to be used in `eql?` Sorry it was a bit ambiguous.

Also, now after looking at the code, the problem initially can be fixed just by adding `eql?` in the abstract proxy. But you additionally suggested that two cells that point at the same actor should also be eql?

ioquatix:
I guess my concern would be that using eql? on two distinct objects which represent the same thing might be confusing, if you had two cells that referred to the same actor and put that in a set they would stomp on top of each other. However, I could see use in defining a custom `==` which does allow this loose form of equality. What do you think? I'll submit two separate PRs.

tarcieri:
@ioquatix would rather discuss `kind_of?` / `is_a?` in a separate issue

ioquatix:
Yeah, so this PR is essentially done. Just allow `eql?` to be used alongside `hash`. Feedback appreciated on this specific issue, and I'll open a new PR where we can discuss other changes to the Proxy stuff.

tarcieri:
This PR looks fine. Feel free to merge on green.

digitalextremist:
:+1:

ioquatix:
Hey everyone who was interested in this issue, could you please review https://github.com/celluloid/celluloid/pull/689 as it may affect you and it would be great to have some feedback.

digitalextremist:
@ioquatix was any branch/commit removed? Travis seems to be choking on the clone step.

ioquatix:
Oh, yeah I deleted the branch since all the tests were green. There are two tasks, one for pr and one for push but they were running the same tests.

PR #689, opened 2015-11-04
Simplified sync, async and future proxies by providing specific AbstractCall base.

ioquatix:
The shared base class additionally implements `eql?` and `hash` which look through to the target mailbox.

I believe this is a good improvement to the structure of the code. We should also check if some improvements could be made to `C::P::Block`

### DIFF
--- a/lib/celluloid/proxy/abstract.rb
+++ b/lib/celluloid/proxy/abstract.rb
@@ -17,3 +17,28 @@ class Celluloid::Proxy::Abstract < BasicObject
     alias_method(:equal?, :==) unless instance_methods.include?(:equal?)
   end
 end
+
+class Celluloid::Proxy::AbstractCall < Celluloid::Proxy::Abstract
+  attr_reader :mailbox
+  
+  def initialize(mailbox, klass)
+    @mailbox = mailbox
+    @klass = klass
+  end
+  
+  def eql?(other)
+    self.__class__.eql?(other.__class__) and @mailbox.eql?(other.mailbox)
+  end
+
+  def hash
+    @mailbox.hash
+  end
+
+  def __klass__
+    @klass
+  end
+
+  def inspect
+    "#<#{self.__class__}(#{@klass})>"
+  end
+end
--- a/lib/celluloid/proxy/abstract.rb
+++ b/lib/celluloid/proxy/abstract.rb
@@ -1,10 +1,14 @@
+module Celluloid::Proxy
+  CLASS_METHOD = ::Kernel.instance_method(:class)
+  
+  # Looks up the actual class of instance, even if instance is a proxy.
+  def self.class_of(instance)
+    CLASS_METHOD.bind(instance).call
+  end
+end
+
 # Base class of Celluloid proxies
 class Celluloid::Proxy::Abstract < BasicObject
-  # Used for reflecting on proxy objects themselves
-  def __class__
-    ::Celluloid::Proxy::Abstract
-  end
-
   # Needed for storing proxies in data structures
   needed = [:object_id, :__id__, :hash, :eql?, :private_methods] - instance_methods
   if needed.any?
@@ -15,6 +19,10 @@ class Celluloid::Proxy::Abstract < BasicObject
     # rubinius bug?  These methods disappear when we include hacked kernel
     define_method :==, ::BasicObject.instance_method(:==) unless instance_methods.include?(:==)
     alias_method(:equal?, :==) unless instance_methods.include?(:equal?)
+  end
+  
+  def __class__
+    @class ||= ::Celluloid::Proxy.class_of(self)
   end
 end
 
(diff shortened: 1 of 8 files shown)


ioquatix:
@tarcieri does this implement your desired behaviour? It does mean that multiple `Cell` instances representing the same actor would clobber each other if put into a hash table.

Additionally, perhaps it makes sense to improve `hash` to take into account more than just the `@mailbox` but also the `__class__`.

Also of note is the addition of `__klass__` which was required for one of the specs. Hopefully this doesn't clobber anything.

lib/celluloid/proxy/abstract.rb
```diff
@@ -17,3 +17,28 @@ def __class__
...
+class Celluloid::Proxy::AbstractCall < Celluloid::Proxy::Abstract
+  attr_reader :mailbox
+  
+  def initialize(mailbox, klass)
+    @mailbox = mailbox
+    @klass = klass
+  end
+  
+  def eql?(other)
+    @mailbox.eql?(other.mailbox) and self.__class__ == other.__class__
```
tarcieri: I would suggest reversing this:

```ruby
self.__class__ == other.__class__ && @mailbox.eql?(other.mailbox)
```

Otherwise you run the risk that `other` does not respond to `#mailbox`

ioquatix: Ah yep good point, thanks!! :)

ioquatix: Also we could blow up if `__class__` doesn't exist...?

tarcieri: If `__class__` doesn't exist all bets are off IMO

digitalextremist:
@ioquatix I am not seeing any invocations of that instance variable across the entire celluvoid. Waiting to see how the tests behave. At first glance, what would be "synchronous" behavior seems to be achieved by invoking `#thread` on the proxy which in turn invokes `#thread` on the `@actor_proxy` ... so at first glance it does seem like a vestigial tail.

ioquatix:
> At first glance, what would be "synchronous" behavior seems to be achieved by invoking #thread on the proxy which in turn invokes #thread on @actor_proxy ... at first glance it does seem like a vestigial tail.

Yes, I did a quick check and came to the same conclusion. It wasn't directly exposed in any meaningful way either.

ioquatix:
There is one concern I still have regarding `eql?` which I've not listed yet.

Do we expect `eql?` to be symmetric?

A similar issue caused problems here, for example: https://github.com/rails/rails/pull/21504 - where == was overloaded by the class so that `object == :symbol` did something different from `:symbol == object`.

We should consider if there are implications. If someone stores something in a hash, and our C::P::Cell instance has the same `hash` code as, say, a `String` instance, would this cause unexpected problems?

IMHO, this may be the best we can do, and then make sure it is recommended that you don't store more than one kind of thing in the `hash` (i.e. Proxy instances or anything else). But, it still bears thinking about this case, because currently it would cause issues.

tarcieri:
@ioquatix I don't see how `eql?` would be asymmetric in this case

ioquatix:
Because most implementations of `eql?` would start with `self.class.eql?(other.class)` which would (from my understanding) look at the class of the actor rather than our implementation which is `self.__class__.eql?(other.__class__)` which would break on anything that doesn't implement `__class__`.

tarcieri:
Adding a check for `respond_to?(:__class__)` (possibly aliasing `__respond_to?` to the corresponding Kernel method to do that) should take care of that

ioquatix:
@tarcieri Yeah that's what I was thinking. But, right now, `respond_to?` would get forwarded to the actor right?

tarcieri:
That's why I was suggesting a `__respond_to?` :confused:

ioquatix:
Maybe it would be better to have something like: `Proxy.is_a?(object)` which does some check which doesn't require adding more methods to the proxy object. I was wondering if the Proxy object should have some non-public way of accessing it's internal structure, e.g. `Proxy.unwrap(object)` which provides some kind of instance which doesn't forward methods to the actor. Maybe it is better to investigate some options which don't require adding lots of `__methods`.

ioquatix:
Is everyone good to merge these changes?

lib/celluloid/proxy/abstract.rb
```diff
@@ -17,3 +17,28 @@ def __class__
...
+class Celluloid::Proxy::AbstractCall < Celluloid::Proxy::Abstract
+  attr_reader :mailbox
+  
+  def initialize(mailbox, klass)
+    @mailbox = mailbox
+    @klass = klass
+  end
+  
+  def eql?(other)
+    self.__class__.eql?(other.__class__) and @mailbox.eql?(other.mailbox)
```
tarcieri: I think this will still crash if `other` doesn't implement `__class__`

ioquatix:
Okay so this is probably an improvement.. what do you think?

ioquatix:
Any idea how we can fix this on JRuby?

ioquatix:
> I hate to say it, but this may actually be one of those times a class variable is actually appropriate

Can you elaborate on this a bit?

Also the proposed solution which also works on JRuby would force a metaclass to be allocated just to figure out the super-class. Not sure it's ideal.. It would be nice if there was a function like `Class.of(instance)` and perhaps `Class#instance_of?`.

ioquatix:
Also wonder if it's worth considering this: http://jakeyesbeck.com/2015/11/08/ruby-delegate-class/

ioquatix:
Unless anyone has any issues I will go ahead and merge this. 48 hours or forever hold your peace :)

tarcieri:
:+1:

ioquatix:
Swee one thumbs up I'm merging

PR #692, opened 2015-11-19
Sidekiq 4.0 and up doesn't have celluloid as a dependency

rebyn:
[ci skip]


tarcieri:
I'd probably prefer to just reword the whole thing rather than call it out like this.

rebyn:
@tarcieri thanks for the feedback. How would you reword the whole thing if I may ask?

digitalextremist:
We just remove it. Now Sidekiq doesn't use Celluloid; done.

PR #695, opened 2015-12-08
extending the condition event handler with the block; this solves a b…

### LINKED ISSUE #694: JRuby block handling messing celluloid event handling
Refering to this line:

https://github.com/celluloid/celluloid/blob/master/lib/celluloid/system_events.rb#L113

This apparently fails in JRuby. I switched this locally to:

```ruby
handler do |e|
  e.call
end
#handler(&:call)
```

and it worked again.

My guess is, this is either a JRuby bug or undefined behaviour. I first thought it was this:

```ruby
 p = proc { puts "balls" }
 p.call #=> "balls"
p.call(1) #=> "balls"

 def test ; puts "balls" ; end 
method(:test).to_proc.call #=> "balls"
method(:test).to_proc.call 1 #=> ArgumentError
```
but CRuby 2.0 behaves the same way. Then I thought it was because of the [define_method meta-definition](https://github.com/celluloid/celluloid/blob/master/lib/celluloid/system_events.rb#L31) that could be different, but I somehow couldn't reproduce the error. This seems, however, to work flawlessly in CRuby but fail all the time in JRuby (at least since 9.0.0.0, according to #677 ). 

@digitalextremist  can you give your 2 cents? (this block seems to come from you). @headius what do you think, is jruby to blame for this? If the problem is indeed jruby, can't we just replace that &:call part with a proper block and event method call, as this is supported? I can submit a pull request as soon as we figure out where should this be fixed.

TiagoCardoso1983:
…ug introduced in jruby >9.0.0.0, which breaks with an ArgumentError exception, apparently due to the way to_proc procs are passed arguments (Fixes #694)

### DIFF
--- a/lib/celluloid/system_events.rb
+++ b/lib/celluloid/system_events.rb
@@ -110,7 +110,9 @@ module Celluloid
     end
     attr_reader :task, :value
 
-    handler(&:call)
+    handler do |event|
+      event.call
+    end
 
     def call
       @task.resume(@value)


TiagoCardoso1983:
@digitalextremist that should do the trick.

digitalextremist:
Very good, thank you @TiagoCardoso1983. Merged :+1:

PR #701, opened 2016-01-11
Conditions in loose threads loop does not take into account the diffe…

### LINKED ISSUE #165: Celluloid::PoolManager crashed!
I'm not sure if i'm doing this right, anyways I'm using ruby 1.9.3, and testing it now on Mac OSX 10.7. 

I have a Sender class which is supposed to send requests to some URL : 

```
require "celluloid"
require "curb"

class Sender
  include Celluloid

  def send(msg)    
    Curl.get($URL) do |url|
      url.headers["content-type"] = "text/html;charset=utf-8"
    end
  rescue Exception => e
    $LOG.error "[erro]-> [ #{e.message} "
  end
end
```

Now I have another process which is supposed to receive msgs from a queue ( RabbitMQ ) and send requests using Sender

```
 AMQP.start(:host => $AMQP_URL) do |connection|
   @channel ||= AMQP::Channel.new(connection)
   @queue   ||= @channel.queue("results")

   pool = Sender.pool

   @queue.subscribe do |metadata, body|
      msg = MessagePack.unpack(body)        
      pool.send(msg)
  end
 end
```

Now when I run it I'm getting this : 

https://gist.github.com/cubiic/31ed2663015a8dbacd2f

Am I don't anything wrong to make the processes crash cuz of the pool manager ?

TiagoCardoso1983:
…rence between backtraces from ruby 2.0.0 and greater than (Fixes https://github.com/celluloid/celluloid-io/issues/165)

### DIFF
--- a/lib/celluloid/rspec.rb
+++ b/lib/celluloid/rspec.rb
@@ -16,7 +16,6 @@ module Specs
 
   CHECK_LOOSE_THREADS = !Nenv.ci? unless defined? CHECK_LOOSE_THREADS
   ALLOW_RETRIES = 3 unless defined? ALLOW_RETRIES
-  ALLOW_SLOW_MAILBOXES = false unless defined? ALLOW_SLOW_MAILBOXES
 
   INCLUDE_SUPPORT = [
     "env",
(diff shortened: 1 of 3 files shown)


ioquatix:
Nice work this is awesome.

PR #702, opened 2016-01-13
Update Ruby versions and add 2.3.0

mlarraz:
.travis.yml hasn't been updated in a while and was testing on 2 different versions of MRI 2.2.x.

I had a problem with a gem after updating to MRI 2.3. Its tests were failing and raising errors from inside Celluloid, so I ran the Celluloid suite and everything passed there. Mostly adding this to save other people time and let them know Celluloid works on MRI 2.3.


.travis.yml
```diff
@@ -3,9 +3,9 @@ language: ruby
 rvm:
   - rbx-2
   - jruby
-  - 2.2.2
-  - 2.2.0
-  - 2.1.4
+  - 2.3.0
+  - 2.2.4
+  - 2.1.7
```
tarcieri: 2.1.8 is out

mlarraz: Whoops, guess I'm not keeping up as much as I thought.

digitalextremist:
Favoring broader tune-up, and included request in f8b5e7c

PR #705, opened 2016-01-26
adding method source code path to backtrace

TiagoCardoso1983:
Opening the discussion here of making celluloid backtraces more useful. 

Usually I have the problem of a method call error, because I'm passing the wrong arguments, and then I get an ugly call stack referring to celluloid internals. 

This commit solves the problem of knowing which method it was about (ArgumentError doesn't specify which method was called with wrong arguments). It doesn't solve the problem of knowing exactly where the method has been called with wrong parameters. One step at a time.

### DIFF
--- a/lib/celluloid/calls.rb
+++ b/lib/celluloid/calls.rb
@@ -45,10 +45,18 @@ module Celluloid
       arity = meth.arity
 
       if arity >= 0
-        fail ArgumentError, "wrong number of arguments (#{@arguments.size} for #{arity})" if @arguments.size != arity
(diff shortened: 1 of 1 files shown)


tarcieri:
Looks good to me :+1:

digitalextremist:
Looked great @TiagoCardoso1983

TiagoCardoso1983:
Thx! will see if I can get to pass around the callstack of the method around to make this more meaningful, but not going to waste much time, as one might have to keep it in memory during method execution.

PR #708, opened 2016-03-14
Added example for using Pools

aayushranaut:
Added example for using Pools, which can be utilized for benchmarking.

### DIFF
--- a/examples/pool.rb
+++ b/examples/pool.rb
@@ -1,0 +1,25 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH.push File.expand_path("../../lib", __FILE__)
+require "celluloid/autostart"
+require "mathn"
+
+#
+# Basic pool example
+#
+
+class PrimeWorker
+  include Celluloid
+
+  def prime(number)
+    puts number if number.prime?
+  end
+end
+
+# Create a pool of PrimeWorker
+pool = PrimeWorker.pool
+
+(2..1000).to_a.map do |i|
+  # Call the prime function asynchronously on the pool
+  pool.async.prime i
+end


chuckremes:
A few comments....

1. If you are trying to use this as a benchmark, I'm certain that most of the CPU time will be consumed by the call to `prime?` in the code. I don't know how that is implemented in the library so it may or may not be a good choice for showing off parallelism.

2. I think it would be useful to allow this example to take a parameter indicating size of pool. I believe the default pool size is 4 (not sure) but that is insufficient to  demonstrate parallelism on a modern desktop that may have 16+ cores.

3. I would suggest adding a comment after the code that shows the results of running this code on your system under MRI, JRuby, and Rubinius.

Looks fine.

digitalextremist:
0. This should be in the `celluloid/celluloid-pool` repository, or `celluloid/celluloid-extras`, fyi.

0. Yes, definitely specify the number of workers... but not just that, I'd actually loop through several sizes... i.e. `1` ( really ) and `2` ... then exactly the number of cores you have, and then double the number of cores... so let's say `1`, `2`, `4`, `8`, and say `16` for good measure ... then loop through those values, running the example/test/benchmark 5+ times.

0. Per @chuckremes' comment, I'd make your own version of `prime?` and start using `benchmark-ips` within that, checking the individual operations rather than just one big black/gray box method.

0. This is my own opinion, but I'd call the `prime` method `check` ... and describe with comments what you expect the behavior to be, because as it stands -- this is not self-documenting code.

0. Do not use `puts` but rather `Celluloid::Internals::Logger` methods, such as `info` since this is going to be spitting out numbers in seemingly random order... with actors otherwise clobbering each other.

digitalextremist:
@aayushranaut please go ahead and move this, after you factor-in the feedback you got... and reference this PR when you make a new one!

PR #709, opened 2016-04-03
Fix autostart.

### LINKED ISSUE #698: Remove default at_exit handler?
Celluloid automatically installs an `at_exit` handler when `celluloid` is required, which just calls `Celluloid.shutdown`. Maybe it shouldn't.

With Rails pulling in Celluloid by default now, this becomes pretty onerous. I think it's contributing to #697.

Celluloid has always straddled the line between library and framework. I think most people end up using it as a library, and for them, the default `at_exit` handler probably isn't helping. Furthermore, the shutdown system has some pretty longstanding problems, and unless it works reliably opting anyone who requires `celluloid` in their app probably isn't a good idea.

I would suggest removing the `at_exit` handler and having people who want the shutdown behavior explicitly call `Celluloid.shutdown`

/cc @dhh

ioquatix:
Fixes #698 - suggestions welcome.

### DIFF
--- a/lib/celluloid.rb
+++ b/lib/celluloid.rb
@@ -531,7 +531,3 @@ Celluloid.group_class =
     end
   end
 
-unless defined?($CELLULOID_TEST) && $CELLULOID_TEST
-  Celluloid.register_shutdown
-  Celluloid.init
-end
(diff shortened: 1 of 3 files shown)


tarcieri:
:+1:

ioquatix:
Dude, you have to use the new Facebook emotes.

PR #710, opened 2016-04-04
Remove deprecated

ioquatix:
This is a work in progress to remove all deprecated code in the main celluloid repository in preparation for a 0.18.0 release and ultimately a library based 1.0.0

Some suggestions/ideas as they come up:
- Move all managed/services/autostart to new gem `celluloid-managed` or something like that.
- Performance evaluation of core method dispatch code.


.travis.yml
```diff
@@ -7,30 +9,21 @@ rvm:
   - 2.2.0
   - 2.1.4
   - 2.0.0
-  - 1.9.3
```
tarcieri: Oh yes please...

ioquatix:
Okay, after reviewing the core celluloid code base I have some major concerns:

- Actor::System is incomplete and badly designed. `running`/`shutdown` are fundamentally broken and the borked behaviour is currently hidden by incorrect exception handling.
- celluloid-essentials needs to be in the main repo, there is a ton of stuff in there that needs to be refactored but it's basically impossible.
- Fundamental issues with how group/system/spawner work tougher.
- Way to much thread local state which is hard to manage, understand and I'm pretty sure there are race conditions.
- Too many configuration options, many of which are remnants of different approaches that are no longer valid or useful.

I think we need a major overhaul of how `Actor::System` works.. I'd like to see the core celluloid merge with the necessary parts of celluloid-essentials and then reduced in scope.

I don't like the way System is a class inside Actor, that implicates a nested dependency, but the dependency is really the other way around.

The use of cells/behaviours/actors and other abstractions are incredibly confusing and it's hard to know what you are dealing with when debugging the code.

digitalextremist:
This is a great start to deprecating old code. I agree `Actor::System` needs to be assessed. Maybe when @chuckremes comes to San Diego in a few weeks we can discuss this, or refactor in the time leading up to that... Agreed also that `celluloid-essentials` ought to be consolidated (again) during the omnirepo merge. Will expect to discuss this again in the next week or two while we come to an arrangement.

ioquatix:
I like the design of Roonga (Groonga). It has a default "context" but you can specify one at any time using options. A top level container which manages actors (`Celluloid::Context`) should be sufficient. Additional logic could be supplied by subclassing this, and if it REALLY makes sense by composing objects. Let's face it, thread allocation in Ruby is not hard, we don't really need different allocation strategies for this until we get the basics right. The defaults should all make sense. Things like fibered vs threaded should either be a decision that's made for you (e.g. implicit in the design of the library) or perhaps an option at actor instantiation. I can imagine both have valid use cases and it shouldn't be an either/or decision, but an option at a fine grained level.

I am going to merge in these changes and then start a major PR for refactoring. I'm currently don't support making an omni-repo. I'd like to merge the internals code directly into `lib/celluloid/internals` and then refactor bits of it to start with.

tarcieri:
Regarding `Actor::System`, it was supposed to be inspired from the same concept from Akka:

http://doc.akka.io/api/akka/2.0/akka/actor/ActorSystem.html

I agree it is not in a good state at present.

ioquatix:
Okay, I'm planning to do a bit more work in this area and I'd like to merge them soonish so we can start progressing forward. Is everyone okay with that?

digitalextremist:
@ioquatix would like to review since there are external pieces I'll need to change too, which is totally fine. Other than that, have at it!

tarcieri:
@ioquatix 👍

ioquatix:
Are you guys okay if I really have at it and try to fix this god awful mess? Like I am happy to have a go at this but you have to be willing for me to break a ton of shit.

tarcieri:
@ioquatix go for it

ioquatix:
Okay, so I MAY eventually revisit this. However, I've been working on something in the mean-time.

https://github.com/socketry/async

This is a super tiny asynchronous reactor implementation for Ruby.

PR #712, opened 2016-04-12
Add RBX-3 to the build

thedrow:


tarcieri:
I ended up reverting this in 8f023f2, although perhaps a bit hastily.

PR #713, opened 2016-04-19
when exit exception is not a SystemExit, then it is some other kind o…

grosser:
…f failure

when trying to use simplecov / single_cov with celluloid I noticed that
adding it makes other at_exit hooks no longer show an error,
which means they can no longer tell if something failed or not /
it always looks as if the exit was 0

see https://github.com/colszowka/simplecov/issues/41

makes this call failed -> also prints coverage when tests failed
https://github.com/grosser/single_cov/blob/master/lib/single_cov.rb#L67

### DIFF
--- a/lib/celluloid.rb
+++ b/lib/celluloid.rb
@@ -169,9 +169,11 @@ module Celluloid
         if defined?(RUBY_ENGINE) && RUBY_ENGINE == "ruby" && RUBY_VERSION >= "1.9"
           # workaround for MRI bug losing exit status in at_exit block
           # http://bugs.ruby-lang.org/issues/5218
-          exit_status = $ERROR_INFO.status if $ERROR_INFO.is_a?(SystemExit)
+          # return previous status or assume something failed
+          # see https://github.com/colszowka/simplecov/issues/41
+          exit_status = ($ERROR_INFO.is_a?(SystemExit) ? $ERROR_INFO.status : 1)
           Celluloid.shutdown
-          exit exit_status if exit_status
+          exit exit_status
         else
           Celluloid.shutdown
         end


tarcieri:
Can we just nuke this whole chunk of code? It looks like it's a workaround for a bug that was fixed years ago:

https://bugs.ruby-lang.org/issues/5218

grosser:
yeah ... removing the logic works ... but only when Celluloid.shutdown is also removed :D 

order is:
```
minitest at_exit #1
minitest at_exit #2 -> exit false
celluloid at_exit -> $! status is 1
single_cov at_exit -> no $!
```

doing a raise+rescue instead of the shutdown make the bug appear too -> not fixed :(

```Ruby
at_exit do
  puts "1 #{$!}"
end

at_exit do
  puts "2 #{$!}"
  raise 'x' rescue nil
end

at_exit do
  puts "3 #{$!}"
  exit 1
end
```

-> a different bug in ruby ... but similar ...

grosser:
ok to merge ?

alternative implementation could be:

```
      at_exit do
        previous_exception = $!
        sleep 0.126 # hax grace period for unnaturally terminating actors
        Celluloid.shutdown
        raise previous_exception if previous_exception
      end
```

basically making the exception handling transparent and not messing with the exit status ...

grosser:
opened issue https://bugs.ruby-lang.org/issues/12302 ... but might end up as wont-fix since rescue/raise is supposed to change $! ...

grosser:
pushed to get a new test run, but now it fails on something different on rbx ...

tarcieri:
Ugh, that looks related to #712, which I'll revert.

grosser:
all good now at least ...

so do you want this PR (what simplecov does) or the re-raise variant (what should be more correct) ?

tarcieri:
I don't know offhand

grosser:
smallest impact fix would be this:

```
    at_exit do
        previous_exception = $!
        sleep 0.126 # hax grace period for unnaturally terminating actors
        Celluloid.shutdown
        raise previous_exception if previous_exception != $!
      end
```

I'd love to get this fixed either way ... picked the simplecov solution (current PR) would be the smallest change ... anyone else I should ping about this ?

grosser:
@ioquatix can I bother you with this issue ... would love to get one of the proposed solutions merged ... just pick one and I'll make the PR ...

ioquatix:
This is an interesting issue relating to the use of Celluloid as a framework rather than a library. It might make sense if we review this usage. I recently merged changes to make celluloid behave more like a library, so unless you require 'celluloid/autostart' you won't get the at_exit block.

ioquatix:
Why can it not just be (from autostart)

```ruby
      at_exit do
        Celluloid.shutdown
      end
```

ioquatix:
Also I'd suggest that using autostart is a hack.

grosser:
Celluloid.shutdown rescues and raises, so the original exception that triggered the exit needs to be reset, see `previous_exception` example above.

I could solve that for my usage by not using autostart and manually doing the re-raise, but I think a solution here would be nicer / cleaner.

grosser:
the autostart merge actually solves my issue since it no longer always installs the exit hook ... but I'd recommend still fixing this issue ... we know what the problem is and how to solve it ... so might as well fix it for good :)

ioquatix:
Perhaps we shouldn't call Celluloid.shutdown if an exception is being thrown?

ioquatix:
e.g.

```ruby
at_exit do
    Celluloid.shutdown unless $!
end
```

grosser:
I'm not sure what shutdown does ... but the process will stop after that
and whatever it does needs to be done no matter if there was an exception
...

ioquatix:
Yeah, but I guess I'm thinking that if the app is being terminated by an exception, there really isn't anything `Celluloid.shutdown` can do except cause chaos. If the user wants to gracefully shutdown they shouldn't be throwing uncaught exceptions. For example, we might be calling Celluloid.shutdown when an out of memory situation occurs, or the disk space is run out, etc. I can't see any situation here where we can reasonably call Celluloid.shutdown automatically without invoking the wrath of undefined behaviour.

grosser:
Then not doing the shutdown would be fine ... unless there is some way of using shutdown to finish work/undo things ... I never used it for anything.

ioquatix:
If anyone ha an issue with the way I've simplified this please feel free to discuss.

Issue #263, opened 2013-05-17
spec directory in gem is incomplete


pravi:
celluloid directory and spec_helper.rb is missing from the gem file. Please add it.

PR #714, opened 2016-04-20
ship less files and remove defaults

grosser:
- makes gem lighter / faster to install and vendor/cache smaller ...
 - remove default values to make gemspec easier to read

... no idea what the culture folder is and why there is a whole other gem in there ideally don't ship that either ...

### DIFF
--- a/celluloid.gemspec
+++ b/celluloid.gemspec
@@ -1,11 +1,8 @@
-# -*- encoding: utf-8 -*-
-
-require File.expand_path("../culture/sync", __FILE__)
+require_relative "../culture/sync"
 
 Gem::Specification.new do |gem|
   gem.name        = "celluloid"
   gem.version     = Celluloid::VERSION
-  gem.platform    = Gem::Platform::RUBY
   gem.summary     = "Actor-based concurrent object framework for Ruby"
   gem.description = "Celluloid enables people to build concurrent programs out of concurrent objects just as easily as they build sequential programs out of sequential objects"
   gem.licenses    = ["MIT"]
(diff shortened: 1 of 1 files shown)


tarcieri:
Including the specs in the gem was a deliberate decision, as downstream Linux distribution packagers have requested them. See #263

grosser:
I don't really care either way, but my arguments in favor of removing them
are:

 - Requesting them does not mean they have to get them
(the 1% asking for a feature that makes the 99% worse off)
 - I don't know anyone using system-packaged ruby gems and I would
recommend against using them in general, so not something I want to
support/encourage.
 - They don't need them to do their packing, works for rails and many others
 - Celluloid is not doing something OS specific / has c-extensions, so
should not break anywhere

tarcieri:
I don't think the size difference is enough to warrant breaking package maintainers.

Celluloid is an infrastructural Ruby library used by many projects. There are Rails apps like Disapora which are apt-get installable and depend on packaged RubyGems, and this would break them.

My vote is in favor of not breaking things over micro-optimizations.

grosser:
Would not break anyone ... just add potential breakage ... but fine either
way.

tarcieri:
It would break downstream packagers when they tried to update the gem

grosser:
they package plenty of gems without tests, so I don't think it would break anything ... unless they have something setup to enforce tests for celluloid

tarcieri:
There are extant downstream maintainers who expect the tests to exist. That's the reality, regardless of what "should be"

PR #715, opened 2016-04-28
fix error response reference in Future#cancel

janx:


digitalextremist:
Thanks for catching this!

PR #717, opened 2016-05-24
Add a clarifying comment to autostart.rb

benzrf:


lib/celluloid/autostart.rb
```diff
@@ -1,5 +1,6 @@
 require "celluloid"
 
+# start up the basic Celluloid services - only necessary to do this manually in old versions
```
tarcieri: `only necessary to do this manually in old versions` not sure what you mean by that

benzrf: do you mean that the phrasing is unclear, or are you asking me to document exactly which version it changed at?

tarcieri: Both would be nice: `celluloid/autostart` was the preferred way to call this for quite some time, so some additional clarification on the history would be good.

digitalextremist: The line in question can also be wrapped in `if $CELLULOID_BACKPORTED`

tarcieri:
Closing as stale

PR #722, opened 2016-06-08
fix $ERROR_INFO warnings

carljohnstone:
When running with warnings enabled I get a warning about; `lib/celluloid.rb:172: warning: global variable '$ERROR_INFO' not initialized`, requiring English fixes this.

### DIFF
--- a/lib/celluloid.rb
+++ b/lib/celluloid.rb
@@ -2,6 +2,7 @@ require "logger"
 require "thread"
 require "timeout"
 require "set"
+require "English"
 
 $CELLULOID_DEBUG = false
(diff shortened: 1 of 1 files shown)


carljohnstone:
better demo of what's happening:
```$ irb -w
2.1.9 :001 > puts $ERROR_INFO
(irb):1: warning: global variable `$ERROR_INFO' not initialized

 => nil 
2.1.9 :002 > require 'English'
 => true 
2.1.9 :003 > puts $ERROR_INFO

 => nil 
```

carljohnstone:
Fixed in #740

PR #727, opened 2016-06-24
Converted all the fail exceptions to raise for convention and testing

prathmeshranaut:
@digitalextremist I've updated all the `fail` exceptions with equivalent `raise` exception.


prathmeshranaut:
@de I've converted all the fail exceptions to raise exceptions and one commit removed the spec helper, which I added back.

PR #729, opened 2016-07-13
Remove mysterious Fiber.yield

tarcieri:
How did this Fiber.yield get here? What is its purpose?

I wrote Celluloid and I can't answer these questions. I don't think it belongs here. I am too lazy to sift through the commit history to find the commit that introduced it.

Per the comment also being deleted, it seems to be causing problems. I've had a few people bug me about it.

Unless there's a justification for it being here, it should be removed.

### DIFF
--- a/lib/celluloid/task/fibered.rb
+++ b/lib/celluloid/task/fibered.rb
@@ -12,8 +12,10 @@ module Celluloid
           Thread.current[:celluloid_queue] = queue
           Thread.current[:celluloid_actor_system] = actor_system
           yield
-          # TODO: Determine why infinite thread leakage happens under jRuby, if `Fiber.yield` is used:
-          Fiber.yield unless RUBY_PLATFORM == "java"
+
+          # Alleged workaround for a MRI memory leak
+          # TODO: validate/confirm this is actually necessary
+          Fiber.yield if RUBY_ENGINE == "ruby"
         end
       end
 


chuckremes:
I read through an old issue (found via `git blame`) that essentially says that MRI will leak fibers without that call to `Fiber.yield`. See here: https://gist.github.com/niamster/862acd8a0c2561940ffc

I do not know if this is still true. Appears it is/was related to MRI's GC not guaranteeing that finalizers run at all except before program exit, so those fibers might not be getting reaped. I'll try and test this out today.

chuckremes:
I tested this out today and confirmed that MRI will not GC the fiber without the call to `Fiber.yield` inside the Fiber block. Neither Rubinius or JRuby need it. Based on comments from @headius in another issue here, he mentioned that calling `Fiber.yield` will actually make the fiber stick around longer on JRuby since the fiber needs another call to `resume` to make it fall out the end of the block.

Honestly, don't know what to do here. Seems like it might be a bug in MRI but my guess is they would just say it isn't deterministic and there is no guarantee the GC will reap the unreferenced fiber if there is no memory pressure.

tarcieri:
Is there any repro for the alleged bug this fixes?

Yielding before the Fiber completes is a pretty nonsensical solution for a memory leak. It's pretty much the opposite behavior I'd expect for a memory leak: I'd expect fibers that don't *complete* are leaked. Yielding immediately before the fiber completes suggests a very deep bug in MRI.

If such a deep bug in MRI exists, and this is really the solution, it should be gated on MRI as the `RUBY_ENGINE` and at very least deserves a comment.

tarcieri:
I've updated the PR to only do this `Fiber.yield` on MRI

PR #752, opened 2017-02-25
Misspellings

olleolleolle:
This trivial PR edits code comments so that the whole project passes a run of the misspellings tool.

([I happened to use Celluloid as an example](http://ollehost.dk/blog/2017/02/25/automate-finding-misspellings-in-source-code/) of where there were a few commonly found misspellings. In order to make things clearer for the reader, and not just be a spelling stickler, I created this PR.)

- In the interest of having 0 misspellings results, I walked over a reference joke to The Man Who Wasn't There in a spec. Perhaps now that the reference joke is destroyed, I should pick a clearer method name ('non_existent_method' for instance)
- All edits affect `git blame` output, of course

### DIFF
--- a/spec/celluloid/calls_spec.rb
+++ b/spec/celluloid/calls_spec.rb
@@ -9,7 +9,7 @@ RSpec.describe Celluloid::Call::Sync, actor_system: :global do
       allow(logger).to receive(:crash).with("Actor crashed!", NoMethodError)
 
       expect do
-        actor.the_method_that_wasnt_there
+        actor.the_method_that_was_not_there
       end.to raise_exception(NoMethodError)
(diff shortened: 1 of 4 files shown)


spec/celluloid/calls_spec.rb
```diff
@@ -9,7 +9,7 @@
       allow(logger).to receive(:crash).with("Actor crashed!", NoMethodError)
 
       expect do
-        actor.the_method_that_wasnt_there
+        actor.the_method_that_was_not_there
```
tarcieri: This breaks the movie reference :cry:

http://www.imdb.com/title/tt0243133/

PR #757, opened 2017-04-18
parameterize error message so old and new Rubies match text

chuckremes:
MRI deprecated the Fixnum class and now expects Integer. So the specs will pass on old and new Rubies alike I just parameterized the error text.


tarcieri:
Thanks!

Issue #696, opened 2015-12-20
"Couldn't cleanly terminate all actors in 10 seconds!" with Spring 1.6.0 and Rails 5.0.0.beta1


dhh:
Rails 5.0.0.beta1 encourages you to mount the Celluloid-powered `ActionCable.server` in-process by changing the routes.rb file to uncomment `mount ActionCable.server => '/cable'`. When you do that, generator commands (that presumably do a fork somewhere) start hanging with the following error message: `E, [2015-12-20T08:09:19.218168 #5720] ERROR -- : Couldn't cleanly terminate all actors in 10 seconds!`.

You can provoke this with:

```
rails new terminator
cd terminator
./bin/rails generate scaffold post title:string
```

If you remove the Spring loader from bin/rails, there is no problem.

dhh:
//cc @tarcieri @grosser

tarcieri:
There was previously a fix for this for 0.16.x (see #612). However, between 0.16 and 0.17 the default thread spawning strategy changed (thread pool -> spawn-on-demand), and the fix for this regressed.

mike-ball:
Any plans to fix this in a new update to 0.17?

ivanilves:
Yes, fix it please :pray:

simonc:
Is there any workaroud in the meantime? It's also happening with a Rails 4 app when Sidekiq is used. Every command takes ages to stop even a simple `rake db:migrate` takes forever :'(

digitalextremist:
@simonc are you using Rubinius or MRI?

simonc:
@digitalextremist MRI v2.3.1.

digitalextremist:
Waiting on CI then good to go.

digitalextremist:
( `listen` and therefore `guard` is causing the build to error, but feel free to use `master` as-is in the meantime; it has the fix )

simonc:
@digitalextremist actually I cannot use `master` right now, when adding this to my Gemfile
```ruby
gem 'celluloid', github: 'celluloid/celluloid'
```
I get the following error when calling `bundle install`
```
Fetching git://github.com/celluloid/celluloid.git
There was a LoadError while loading celluloid.gemspec:
cannot load such file -- /Users/xxx/.gem/ruby/2.3.1/bundler/gems/celluloid-8d8f1d460d79/culture/sync from
[...]

Issue #612, opened 2015-05-05
add fork() fix to 0.18.0 ( scrapped from 0.16.1 )


e2:
@digitalextremist - could you release 0.16.1 with #513 ? Basically Celluloid terminates with an error without it, and it's already causing problems with multiple gems using fork().

e2:
NOTE1: This patch is not on 0-16-stable (it's been reverted)
NOTE2: this patch should be applied to the Spawner as well

digitalextremist:
Where was it reverted?

e2:
Here: https://github.com/celluloid/celluloid/commit/6226a68cd6adf63440c70bcb7a1d7ed68f85ce87

(The revert was merged, not the patch).

digitalextremist:
Alright, I see this and don't see a problem putting the patch into `0.16.1`

e2:
@digitalextremist - any update on this?

digitalextremist:
@e2 I cherry picked that for you, and cut `0.16.1` just now. If you would be willing to reapply the patch to `0.18.0` in its current incarnation and PR that, I'd appreciate it. Otherwise I'll eventually remember to do that, but not for a long time. Please confirm your situation is sorted out now.

e2:
Listen doesn't use celluloid since 3.x, so I personally am not experiencing this anymore - and so I don't have a celluloid project to test a PR against. 

It's unlikely this problem will reoccur, and if it does, it isn't complex to fix with an similar patch.

sebyx07:
-v '0.16.1' is yanked, can you revert this ?

digitalextremist:
Yes, this is yanked. And the piece of logic in this will come back in at `0.18.0` for those following the path forward. Maintaining the previous series of versions has proven to be like trying to brush removed molars, so that'll just stop. If people want what was `0.16.1` then they can pull in `0-16-stable` directly from the repository... but `0.18.0` ought to dramatically out-perform the `0.16.0` line, as the `0.17.0` line already does in many different ways. Whatever ways it lacks currently, will resolved in `0.18.0`

bluej100:
The most recent version of sidekiq still depends on the 0.16 branch: https://rubygems.org/gems/sidekiq/versions/3.4.2 .

digitalextremist:
Yes, and it was just updated to use `0.17.*` over the last 7-10 days, so that'll be nice to see it released.

bluej100:
Terrific. Thank you.

[...]

PR #763, opened 2017-07-14
Bring back fix for #696 from #612, in new ::Group::* convention.

zsiddique:
So this is a patch to make the fix from #696 and apply it to the 0-17-stable branch so a 0.17.4 release can be made.  

This fix has been in the 0.18pre branch for a while.  This would be easier than trying to resolve all the open issue in 0.18 branch.

### DIFF
--- a/lib/celluloid/group.rb
+++ b/lib/celluloid/group.rb
@@ -3,6 +3,7 @@ module Celluloid
     attr_accessor :group
 
     def initialize
+      @pid = $$
       @mutex = Mutex.new
       @group = []
       @running = true
@@ -25,7 +26,12 @@ module Celluloid
       to_a.each { |thread| yield thread }
     end
 
+    def forked?
+      @pid != $$
+    end
+
     def to_a
+      return [] if forked?
       res = nil
       @mutex.synchronize { res = @group.dup }
       res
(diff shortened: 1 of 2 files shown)


zsiddique:
The test failures are due to ruby_dep requiring ruby 2.2 and greater but we have some test running against older ruby version.  The one test passed was against 2.2.5.  I think the fix for that is out of this scope

PikachuEXE:
This very old build on 0.17 branch looks green
https://travis-ci.org/celluloid/celluloid/builds/103210007

But not sure what happens when re-run with latest

tarcieri:
I restarted the build

PikachuEXE:
```
$ rvm use jruby --install --binary --fuzzy
jruby is not installed - installing.
Searching for binary rubies, this might take some time.
Requested binary installation but no rubies are available to download, consider skipping --binary flag.
Gemset '' does not exist, 'rvm jruby do rvm gemset create ' first, or append '--create'.
The command "rvm use jruby --install --binary --fuzzy" failed and exited with 2 during .
```
Not sure Travis issue or Travis config issue

PikachuEXE:
It seems `master` is using a different version of `jruby`
https://travis-ci.org/celluloid/celluloid/builds/299878222

myers:
@tarcieri could you retry the build.  It appears rvm failed to find both rbx and jruby in the last attempt.

I would really like to see a 0.17.4 as it would solve this long yak shave trip I've been on.

tarcieri:
@myers still can't find `rbx-2`. I'm ok with you deleting it, but if the tests pass it'd be nice to have.

Issue #758, opened 2017-05-03
Consecutive Celluloid.boot / Celluloid.shutdown calls throw Celluloid::Supervision::Configuration::Error::InvalidSupervisor


dlau:
Hi,

In specs, I am calling `Celluloid.boot` and `Celluloid.shutdown` before and after each block evaluation, per recommended usage in the wiki.

Gemfile:
```rb
# frozen_string_literal: true
source "https://rubygems.org"

gem "celluloid", "0.18.0.pre"
gem "rspec"

```

Here is a stripped down test file:

```rb
require 'rspec'
require 'celluloid/current'

RSpec.describe 'foo' do
  before(:each) do
    Celluloid.boot
  end

  after(:each) do
    Celluloid.shutdown
  end

  it '1' do
    expect(1).to eq(1)
  end
  it '2' do
    expect(2).to eq(2)
  end
end
```

When run, the output looks like this:
```sh
~/dev/celluloid-test rake                                                                                                                                    
Failures:

  1) foo 2
     Failure/Error: Celluloid.boot

     Celluloid::Supervision::Configuration::Error::InvalidSupervisor:
       Celluloid::Supervision::Configuration::Error::InvalidSupervisor
     # /Users/daryllau/.rvm/gems/ruby-2.3.1/gems/celluloid-supervision-0.20.6/lib/celluloid/supervision/configuration.rb:58:in `provider'
     # /Users/daryllau/.rvm/gems/ruby-2.3.1/gems/celluloid-supervision-0.20.6/lib/celluloid/supervision/configuration.rb:52:in `provider'
     # /Users/daryllau/.rvm/gems/ruby-2.3.1/gems/celluloid-supervision-0.20.6/lib/celluloid/supervision/configuration.rb:65:in `block in deploy'
     # /Users/daryllau/.rvm/gems/ruby-2.3.1/gems/celluloid-supervision-0.20.6/lib/celluloid/supervision/configuration.rb:64:in `each'
     # /Users/daryllau/.rvm/gems/ruby-2.3.1/gems/celluloid-supervision-0.20.6/lib/celluloid/supervision/configuration.rb:64:in `deploy'
     # /Users/daryllau/.rvm/gems/ruby-2.3.1/gems/celluloid-0.18.0.pre/lib/celluloid/actor/system.rb:61:in `block in start'
[...]

PR #766, opened 2017-09-12
Fix celluloid/celluloid#758

sobrinho:


sobrinho:
@tarcieri can you make sure that's the right solution?

tarcieri:
It's definitely a "good enough for now" solution... the API is still not ergonomic, unfortunately

PR #767, opened 2017-09-14
CHANGES: Add GitHub compare link on each heading

olleolleolle:
This PR adds **a GitHub compare link** (which compares last release and current) to each version heading in **the changelog**.

I was inspired by Keep a Changelog, which does this.

Upside: improved metadata.

Downside: one more thing to do, when releasing.


tarcieri:
Neat, thanks!

PR #773, opened 2018-03-20
Bring back fix for #696 from #612, in new ::Group::* convention. (take 2)

myers:
This was proposed in PR #763, but we are stuck with rbx-2 not being installable via rvm in travis.

I'm attempting to use the prebuilt binaries that travis says they have here:

http://rubies.travis-ci.org/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 script: rake ci
 language: ruby
 rvm:
-  - rbx-2
   - jruby
   - 2.2.2
   - 2.2.0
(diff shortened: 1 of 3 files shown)


tarcieri:
@myers "something else" didn't work... just delete rbx :cry:

myers:
@tarcieri please let me know if there is anything I can to assist in seeing a 0.17.4 release.

tarcieri:
@myers the main thing that would be helpful is putting together a changelog entry in [CHANGES.md](https://github.com/celluloid/celluloid/blob/0-17-stable/CHANGES.md)

tarcieri:
All right, looking at the history everything interesting is happening on `master`

PR #787, opened 2018-12-28
Travis config: drop old configuration sudo: false

olleolleolle:
This PR drops an unused Travis configuration line.

See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration


tarcieri:
Thanks again! Cleaning up that cruft 😉

Issue #742, opened 2016-12-05
Unable to enable debug logging without $CELLULOID_DEBUG = true


mconigliaro:
I tried this:

    Celluloid.logger.level = ::Logger::DEBUG

I also tried with my own logger:

    Celluloid.logger = ::Logger.new(STDERR)
    Celluloid.logger.level = ::Logger::DEBUG

But messages logged with `debug()` never appear. The only way I've been able to get debug messages is by setting `$CELLULOID_DEBUG = true`. Is this the intended behavior?

mconigliaro:
I just discovered that calling `Celluloid.logger.debug()` instead of just `debug()` seems to work as expected.

tarcieri:
Ugh, I should just get rid of those globals. So confusing...

andy-twosticks:
Similar problem here -- I have a Logger object, but if I set Celluloid.logger to the logger object, it doesn't output anything.

tarcieri:
Closing as stale. If you are still interested in this issue, please reopen it.

PR #803, opened 2020-06-13
Remove global variable $CELLULOID_DEBUG

emesepadanyi:
With removing this global variable, the logging is more straight forward. 
Setting the level of logging on `Celluloid.logger` will take effect instead of manipulating global variables. The mechanism is the same as specified in [Logger.](https://ruby-doc.org/stdlib-2.2.0/libdoc/logger/rdoc/Logger.html)

Fix for issue #742 .


emesepadanyi:
Yeah, I updated the diagram to these changes, and that is why the files are here.

Celluloid.logger is expected to be a [Logger](https://ruby-doc.org/stdlib-2.2.0/libdoc/logger/rdoc/Logger.html), or nil. This check is already implemented in that class, so it doesn't make sense to double-check.

lib/celluloid/group.rb
```diff
@@ -16,7 +16,7 @@ def assert_active
     def assert_inactive
       return unless active?
       if RUBY_PLATFORM == "java"
-        Celluloid.logger.warn "Group is still active"
+        Internals::Logger.warn "Group is still active"
```
digitalextremist, line 19: @emesepadanyi is this intentionally going around the logger? Am on phone app and cannot reference full source well atm: is `.logger` not initialized here?

digitalextremist, line 19: It seems you are following conventions here, from further edits I see?

emesepadanyi: Yes. As far as I can understand, `Celluloid.logger` can be set to `nil` by the users, in case they don't want any log messages. `Internals::Logger` is sort of a wrapper to catch if `Celluloid.logger` would be nil. Anywhere else in the code `Internals::Logger` is used to trigger the logging, only this place where it is different. I think it might have been an oversight. 
That is why I think it should be like this.

dilumn:
@emesepadanyi I will get `Internals::Logger.debug "Discarded message (mailbox is dead): #{message}"` as an example to explain my question

Previous implementation was
```
Internals::Logger.debug "Discarded message (mailbox is dead): #{message}" if $CELLULOID_DEBUG
```

So we only log `Discarded message (mailbox is dead): #{message}` if ` $CELLULOID_DEBUG` is true. But now there is no more ` $CELLULOID_DEBUG`. So we will log `Discarded message (mailbox is dead): #{message}` every time.

Will it be unnecessary logs because we don't check any condition now?

emesepadanyi:
@dilumn It is true that the variable is not checked anymore, and `Internals::Logger.debug` will get called every time. However, under the hood, the Logger object that actually gets called checks the `Logger.level`. According to the level it logs or discards the message. 
So for example, if `Logger.level = Logger::WARN`, then only warn, error and fatal messages get logged. The order is the following:
DEBUG < INFO < WARN < ERROR < FATAL < UNKNOWN
So in order to get the debug messages, `Celluloid.logger.level` must be set to `Logger::DEBUG`.
I hope it is more clear now, but please refer to the original documentation:
https://ruby-doc.org/stdlib-2.2.0/libdoc/logger/rdoc/Logger.html

Side note:
if the code is:
```
Internals::Logger.debug "Discarded message (mailbox is dead): #{message}" if $CELLULOID_DEBUG
```
It only gets logged if both of these lines are included:
```
Celluloid.logger.level = Logger::DEBUG
$CELLULOD_DEBUG = true
```

spec/celluloid/internals/logger_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+        it { is_expected.to output.to_stderr_from_any_process }
+      end
+
+      context "logging level is INFO" do
+        before do
+          Celluloid.logger = Logger.new(STDERR)
+          Celluloid.logger.level = Logger::INFO
+        end
+        it { is_expected.not_to output.to_stderr_from_any_process }
+      end
```
emesepadanyi, line 35: @dilumn I made some tests to showcase that the level of the logger is enough to manipulate the log messages.

bogdanRada: @emesepadanyi  , i am not a contributor to this project, but my personal thoughts about this is that i think the reason for the global variable to exist was for performance. Since the logger is called every time now, the performance will decrease. Can we benchmark somehow what exactly is the performance difference? That would be great.

Perhaps instead of using a global variable we could use a `mattr_accessor :debug_enabled` on Celluloid and use that class variable. Which should preserve the existing performance. 

This is just an idea, might not be the best solution.

pboling (approved):
LGTM

spec/celluloid/internals/logger_spec.rb
pboling: This is a decent idea, but setting the logger to nil has almost the same effect.  The performance difference between checking a `debug_enabled` and the existing check if there is a logger set, shouldn't be great.

The one thing that does make a difference is when called the logger the string must be computed as an argument, and this can create wasted overhead if the logger isn't even going to be used.  The `Logger` class in Ruby solves for this by allowing it to be called with a block that returns a string.  The block would then only be evaluated if the logger & level would result in printing the log message.  Otherwise the block isn't evaluated.

Switching log statements that are expensive computationally, or that have interpolation, would improve performance slightly.

PR #806, opened 2020-08-28
Add an example and stub function for pool

p-vinayak:
Provides a stub function which returns and error if the user is trying to invoke pool without requiring it.

Also added an example to demonstrate how pool works.


harsh183:
I really like this. I think this is a good resolution to the pool errors that we ran into earlier and all the build errors are resolved as well. I think this is good to merge.

dilumn:
@p-vinayak do you know why this pull request include `examples/configurations.rb` & `examples/supervisors_and_registry.rb` files?
And why github shows them as empty files?

Have you updated your fork & `p-vinayak-pool-example-and-stub-function` branch already with Celluloid master?

p-vinayak:
I added executable permissions for those scripts on my machine because the tests have build errors since it was unable to run them. I got that same error locally and on travis CI.

I haven't updated the celluloid master branch on my fork yet.

examples/pool.rb
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH.push File.expand_path("../lib", __dir__)
+require "celluloid/autostart"
+
+# This is required for pool to work as Celluloid doesn't load it by default
+require "celluloid/pool"
+require "prime"
```
dilumn, line 8: shall we have empty line here?

```
# This is required for pool to work as Celluloid doesn't load it by default
require "celluloid/pool"

require "prime"
```
Just a very small improvement, not doing it also fine

dilumn (approved):
Change here looks good to me. It is good that we added an example for future users to reference

examples/pool.rb
p-vinayak: Sure thing! Will push it soon

PR #809, opened 2020-12-05
v0.18.0 release note

dilumn:
v0.18.0 release note with all merged pull requests to master after `v0.17.3`
https://github.com/celluloid/celluloid/compare/v0.17.3...v0.18.0

`v0.18.0.pre2` released more than two years ago. I think we can release `v0.18.0` & address any bugs as soon as possible. So it will be better for the community who are still using celluloid & development we are trying to ramp up again

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,150 @@
-## 0.18.0.pre2 (Final release TBD)
+## 0.18.0 (TBD)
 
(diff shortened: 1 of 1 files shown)


harsh183:
I think the time is right. Honestly I think `celluloid` is at a pretty stable place right now. We can push out a 0.5 or 1.0 in a few months as well.

PR #818, opened 2023-01-09
[jruby] do not rely on JRuby's Thread internals

kares:
as the Thread#to_java behavior might change in the future,

scripting Java to obtain the current thread is future proof and backwards compatible


ioquatix (approved):
LGTM.

